PR #42, opened 2023-06-01
Fix incorrect wp-cli path for test environment

wojtekn:
In this PR, I propose to fix a bug in a function that gets a path to `wp-cli.phar`.

Currently, for the non-test environment it provides a correct path pointing to `wp-cli.phar` under the `.wp-now` directory, but for the test environment, it returns the temp directory and not the path to the `wp-cli.phar`.


sejas (approved):
@wojtekn , thanks for finding this bug 👍

PR #43, opened 2023-06-01
wp-now: improve php tests and comment failing tests

sejas:
### What?

Fix tests.
Fix Mac Os wp-cli path error.
Fix php tests after playground mode was introduced.

### Testing instrucitons

- Run `npx nx test wp-now`
- Tests should pass.

### DIFF
--- a/packages/wp-now/src/execute-wp-cli.ts
+++ b/packages/wp-now/src/execute-wp-cli.ts
@@ -4,6 +4,7 @@ import { disableOutput } from './output';
 import getWpCliPath from './get-wp-cli-path';
 import getWpNowConfig from './config';
 import { DEFAULT_PHP_VERSION, DEFAULT_WORDPRESS_VERSION } from './constants';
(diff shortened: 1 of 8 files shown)


wojtekn:
@sejas I've tried tests locally and they went successfully.

I tried uncommenting those from `execute-php.spec.ts` and now they still work fine. It crashes for me only when I uncomment 'wp-cli command' tests.

wojtekn (approved):
Great finding @sejas ! We may remove two unnecessary imports from `packages/wp-now/src/tests/execute-php.spec.ts` and should be good to go.

sejas:
@wojtekn Thanks for reviewing it!

I moved some php tests to `wp-now.spec.ts` and deleted the unnecessary imports.

The issue we had was that after the `playground` mode was introduced, the php tests competed with `wp-now.spec.ts` tests to download WordPress returning random errors.

Issue #7, opened 2023-05-24
wp-now requires nvm to run


dmsnell:
After building `wp-now` I attempted to run it, but it failed because a dependency on `nvm` has apparently bled into the project, and since I don't have `nvm` and don't want `nvm` the `bin/wp-now` command fails, which should not happen.

A fix is when running a "binary" doesn't depend on a third-party development version manager.

danielbachhuber:
@dmsnell Do you mean this binary? https://github.com/WordPress/wordpress-playground/blob/trunk/bin/wp-now

adamziel:
Yes

wojtekn:
We have `nvm` listed as a requirement for the project in [CONTRIBUTING](https://github.com/WordPress/wordpress-playground/blob/trunk/packages/wp-now/CONTRIBUTING.md).md docs, so technically, the project depends on it. However, it would be better if running such an executable file wouldn't change the developer's environment.

@danielbachhuber, as we recommend running `nvm use` as part of development setup steps, do we need those in the `bin/wp-now` file?

```
export NVM_DIR=$HOME/.nvm;
source $NVM_DIR/nvm.sh;
nvm use
```

If we remove them, the linked wp-now would be just responsible for running `nx preview wp-now` from the `wordpress-playground`, pretending it runs from the project directory, and wouldn't modify the developer's environment in any way. 

As far as I see, we would still be able to link and run `wp-now` globally, and the command would work for developers who don't want to use nvm.

PR #44, opened 2023-06-01
wp-now: Remove unnecessary nvm calls

wojtekn:
## What?

In this diff, I propose to remove unnecessary nvm calls from the script that allows running wp-now for development purposes globally.

## Why?

- remove dependency on nvm from the code
- ensure that the script located in `bin/` doesn't make changes on the machine when it runs

Related ticket: https://github.com/WordPress/playground-tools/issues/7

## How?

I removed the affected code. We still recommend running `nvm use` in contribution docs, but the user can skip that if wanted to manage node versions in another way.

## Testing Instructions

1. Link wp-now in development mode
2. Ensure wp-now still works when used globally from the plugin/theme or other project root

### DIFF
--- a/bin/wp-now
+++ b/bin/wp-now
@@ -13,7 +13,5 @@ if [[ "$@" != *"--path="* ]]; then
 fi
 
 cd $SCRIPT_DIR/..
-export NVM_DIR=$HOME/.nvm;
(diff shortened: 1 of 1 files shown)


adamziel (approved):
LGTM

Issue #39, opened 2023-05-30
Issues using WordPress Playground VSCode extension with Learn WordPress project


adamwoodnz:
I've been testing the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=WordPressPlayground.wordpress-playground) with the [Learn.WordPress.org](https://github.com/WordPress/Learn) project repo.

Context: https://wordpress.slack.com/archives/C04EWKGDJ0K/p1684877322935099?thread_ts=1684852995.310519&cid=C04EWKGDJ0K

This repo has a custom plugin and child theme, so running from the project root would be the ideal local dev scenario, and this is how we currently run the project with `wp-env`.

The things I tried were:

## Setup
Fresh clone of https://github.com/WordPress/Learn
Basic install of dependencies and build: `yarn`, `composer install`, `yarn workspaces run build`

### Try full project:
Click Start WordPress Server button from project root.

Result: Notification that server started but browser never loads site.

### Try Theme only:
Click Start WordPress Server button from theme folder (https://github.com/WordPress/Learn/tree/trunk/wp-content/themes/pub/wporg-learn-2020)

Result: Notification that server started, browser loads site but TT3 theme is active because parent theme isn’t available.

### Try Plugin only:
Click Start WordPress Server button from plugin folder (https://github.com/WordPress/Learn/tree/trunk/wp-content/plugins/wporg-learn)

Result: Notification that server started but browser never loads site.

## Environment: About VSCode

Version: 1.78.2 (Universal)
Commit: b3e4e68a0bc097f0ae7907b217c1119af9e03435
Date: 2023-05-10T14:44:45.204Z
Electron: 22.5.2
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin arm64 21.6.0
Sandboxed: No

danielbachhuber:
I'll look into this. I think we need to refresh the VS Code Extension's internals (https://github.com/WordPress/playground-tools/issues/4) to get closer to the behavior you expect. It would also be nice to add some debugging mechanism for the crashes (https://github.com/WordPress/playground-tools/issues/36).

> Result: Notification that server started, browser loads site but TT3 theme is active because parent theme isn’t available.

@adamwoodnz Out of curiosity, what is your expected behavior for this scenario? Related https://github.com/WordPress/playground-tools/issues/33

adamwoodnz:
[...]

Issue #26, opened 2023-05-10
Replace "Mode" concept with auto-loading or explicit CLI flags


dmsnell:
`mode` seems arbitrarily limiting as a loading strategy, and as WordPress/wordpress-playground#277 evidences, will need to be routinely expanded to encompass the multitude of ways people have unique setups.

I propose that we replace this concept that we created for this tool with one of two options:
 - automatically infer directories to mount from the location the tool runs from (I have taken this approach in the VSCode extension)
 - provide explicit flags on the CLI to mount the different directories

In the one case there's nothing someone needs to do other than start `wp-cli`. The startup looks for WordPress core files, and takes a guess (e.g. is there a `wp-config.php`). It looks for a `themes` subdirectory or a `plugins` subdirectory and can load either of those or both of those. If there are other subdirectories, it can look for a `.php` or `.css` file that appears to be a plugin or theme and loads those subdirectories accordingly.

Things can go wrong and the Playground can guess wrong, but this keeps the configuration simple, and those failures may not be very big in the grand scheme of things.

In the second case everything is explicit. `wp-cli --core=wordpress-6.2 --plugins=my-plugins --plugin=test-plugin --plugin=another-plugin --theme=parent-theme --theme=child-theme`

adamziel:
I really like this idea, especially since I've already seen a bunch of instances of "it inferred the wrong mode and gave me confusing errors." The "mount everything" mode would make it "just work" with VS Code workspaces where I can have multiple types of WordPress code open as different folders in the same workspace (e.g. themes, plugins, mu-plugins). 

The explicit flags would be a good addition, too, if the automated mode ends up being not enough.

cc @danielbachhuber @sejas

danielbachhuber:
Let's proceed with `mode` for now, and await further feedback from the CfT.

Here are a few opinions I have:
1. We want to optimize for "it's so fast and it just works".
2. We don't want to do a major refactor a few days before opening it up to broader testing.
3. When we do refactor, we'll want to make sure we add test coverage first so we don't break any existing behaviors.

`mode` seems to cover most use cases at the moment, so I think it's good enough for now.

wojtekn:
If we go that way, I would also like to keep the "mount as is" model to support projects that consist of full WordPress installation stored in the project root.

danielbachhuber:
A great (and important, I think) place to start would be to document these various permutations we want to support.

sejas:
[...]

Issue #32, opened 2023-05-12
Avoid adding empty directories in 'wp-content' and 'wordpress' modes


danielbachhuber:
If you run `wp-now start` in 'wp-content' mode, you will end up with some empty directories: `wp-content/database`, `wp-content/plugins/sqlite-database-integration`, and `wp-content/db.php`.

From https://github.com/WordPress/wordpress-playground/pull/318:

> Note: `mount()` still adds an empty directory (https://emscripten.org/docs/api_reference/Filesystem-API.html#FS.mount) but at least such empty directory is not caught by `git status` in the developer's project.

This seems like an upstream bug right now; not sure how we'll be able to resolve.

adamziel:
Sqlite plugin and db.php could perhaps be somehow handled by VFS, but I’d like the database file to stick around. It’s your data, you should have access to it.

wojtekn:
> Sqlite plugin and db.php could perhaps be somehow handled by VFS, but I’d like the database file to stick around. It’s your data, you should have access to it.

I think that keeping it in the hidden directory by default will be a preferable default approach as that way, we will avoid making changes to local project files. As a user, when I clone the repository and start the local environment there, I would expect to have no (or minimal) changes done to my project, so `git status` still shows clean. I implemented it in https://github.com/WordPress/wordpress-playground/issues/313 

We could add the behavior you proposed as an option in the feature, so the user could decide to keep the SQLite database file with the project and track it in VCS, for example.

This ticket's goal is to explore the possibility of removing the side effect of FS.mount, which currently adds an empty directory for each mounted resource.

fmfernandes:
I'm using Laravel Valet to develop locally and trying out `wp-now` as well. 

On a standard installation of WordPress in a parked domain in Valet if I later run `wp-now start` in `wp-content`, the Valet site stops working, throwing a fatal:

```bash
Warning: require_once(/test/wp-content/db.php): failed to open stream: No such file or directory in /test/wp-includes/load.php on line 570

Fatal error: require_once(): Failed opening required '/test/wp-content/db.php' (include_path='.:/opt/homebrew/Cellar/php@7.4/7.4.33_2/share/php@7.4/pear') in /test/wp-includes/load.php on line 570
```

So, I have to manually delete the `database` and `db.php` directories created by `wp-now` for the Valet site to work again. While I'm just testing both tools, it would be good if there's a way to have both working side-by-side.

danielbachhuber:
[...]

Issue #19, opened 2023-05-16
Allow a configuration to map other folders and files


renatho:
It would be nice if there was a way to map files and folders while running a project.

I have this, for example:

```
- woocommerce
    - (plugin) woocommerce
    - (plugin) woocommerce-subscriptions
    - (plugin) woocommerce-memberships
- shared
    - (plugin) query-monitor
    - (plugin) wp-crontrol
- sensei
    - (plugin) sensei-lms
    - (plugin) sensei-pro
- job-manager
    - (plugin) wp-job-manager
    - (plugin) wp-job-manager-applications
    - (plugin) wp-job-manager-alerts
```

Basically, I run 2 different environments (a Sensei and a Job Manager site). And both share plugins and themes from other folders (like the plugins I have in `woocommerce` and `shared` folders).

Currently, I’m doing it with Local using symbolic links. I’d love to have it in a config file, or something like that.

danielbachhuber:
This would be cool 👍 Thanks for the suggestion!

We're adding some limited `wp-env.json` support with https://github.com/WordPress/wordpress-playground/issues/230

We're also considering some auto-setup functionality with https://github.com/WordPress/playground-tools/issues/20

PR #51, opened 2023-06-01
wp-now: Treat `wp-content` in a subdirectory as 'wp-content' mode

danielbachhuber:
See https://github.com/WordPress/playground-tools/issues/39#issuecomment-1572207645
Related https://github.com/WordPress/playground-tools/issues/26

## What?

Treats a repository structure like https://github.com/WordPress/Learn as 'wp-content' mode:

image

Specifically, in the Learn repository, the `wp-content` directory is a subdirectory of the repository, and contains `plugins` and `themes` directories.

## Why?

* We want to support the Learn repository structure.
* We don't want to introduce a new mode.
* Our 'wordpress' mode is described as _Runs the directory as a WordPress installation when WordPress files are detected_, so it's not appropriate.

## How?

* Adapts `isWpContentDirectory()` to first look in the immediate directory for `plugins` or `themes`, and then also look in a `wp-content` subdirectory (if one exists) for `plugins` or `themes` too.
* Updates `runWpContentMode()` to handle both variations of directory structures.

## Testing Instructions

1. Set up the WordPress/Learn repository:

```
git clone git@github.com:WordPress/Learn.git wordpress-learn
cd wordpress-learn
yarn
composer install
yarn workspaces run build
```

2. Run `nx preview wp-now start --path=/path/to/learn-repository`
3. Verify plugins and themes are mounted as expected.
4. Test `nx preview wp-now start` against a few other modes and verify it works as expected.

### DIFF
--- a/packages/wp-now/src/wp-now.ts
+++ b/packages/wp-now/src/wp-now.ts
@@ -2,6 +2,7 @@ import fs from 'fs-extra';
 import { NodePHP, PHPLoaderOptions } from '@php-wasm/node';
 import path from 'path';
 import { SQLITE_FILENAME } from './constants';
+import { hasPluginOrThemeDirectory } from './wp-playground-wordpress/has-plugin-or-theme-directory';
 import {
 	downloadMuPlugins,
 	downloadSqliteIntegrationPlugin,
@@ -163,7 +164,11 @@ async function runWpContentMode(
 	await initWordPress(php, wordPressVersion, documentRoot, absoluteUrl);
 	fs.ensureDirSync(wpContentPath);
 
-	php.mount(projectPath, `${documentRoot}/wp-content`);
+	if (hasPluginOrThemeDirectory(projectPath)) {
+		php.mount(projectPath, `${documentRoot}/wp-content`);
+	} else {
+		php.mount(projectPath, documentRoot);
+	}
 
 	mountSqlitePlugin(php, documentRoot);
 	mountSqliteDatabaseDirectory(php, documentRoot, wpContentPath);
--- a/packages/wp-now/src/wp-now.ts
+++ b/packages/wp-now/src/wp-now.ts
@@ -167,7 +167,7 @@ async function runWpContentMode(
 	if (hasPluginOrThemeDirectory(projectPath)) {
 		php.mount(projectPath, `${documentRoot}/wp-content`);
 	} else {
-		php.mount(projectPath, documentRoot);
+		php.mount(`${projectPath}/wp-content`, `${documentRoot}/wp-content`);
 	}
 
 	mountSqlitePlugin(php, documentRoot);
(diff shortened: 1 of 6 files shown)


packages/wp-now/src/wp-now.ts
```diff
@@ -163,7 +164,11 @@ async function runWpContentMode(
 	await initWordPress(php, wordPressVersion, documentRoot, absoluteUrl);
 	fs.ensureDirSync(wpContentPath);
 
-	php.mount(projectPath, `${documentRoot}/wp-content`);
+	if (hasPluginOrThemeDirectory(projectPath)) {
+		php.mount(projectPath, `${documentRoot}/wp-content`);
+	} else {
+		php.mount(projectPath, documentRoot);
+	}
```
danielbachhuber, line 171: @sejas Any ideas why the Learn `wp-content` directory isn't being mounted correctly?

image

danielbachhuber: As it turns out, it was following the first part of the conditional because there were some hidden directories (https://github.com/WordPress/playground-tools/issues/32):

image

The hidden directories existed because I previously ran `wp-now start` in the directory.

When I remove the hidden directories, I see this crash:

```
Error: Could not mount /Users/danielbachhuber/projects/wordpress-learn: Device or resource busy.
    at _NodePHP.descriptor.value (/Users/danielbachhuber/projects/playground-tools/node_modules/@php-wasm/node/index.cjs:67405:17)
    at runWpContentMode (file:///Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/main.js:590:9)
    at async file:///Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/main.js:535:9
    at async applyToInstances (file:///Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/main.js:481:5)
    at async startWPNow (file:///Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/main.js:532:3)
    at async startServer (file:///Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/main.js:780:42)
    at async Object.handler (file:///Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/main.js:925:25) {
```

However, if `wp-now` creates hidden directories each time it starts, we'll revert to the first behavior after it starts...

danielbachhuber: With more reflection, I kinda think this particular scenario would be better solved by https://github.com/WordPress/playground-tools/issues/19

sejas: > @sejas Any ideas why the Learn `wp-content` directory isn't being mounted correctly?
@danielbachhuber , I debugged the code and `php.mount(projectPath, documentRoot);` is trying to mount twice the same destination`documentRoot`. This directory was already mounted with the WordPress latest files.

A simple suggestion would be mounting the `wp-content` folders.
```diff
- php.mount(projectPath, documentRoot);
+ php.mount(`${projectPath}/wp-content`, `${documentRoot}/wp-content`);
```
But this approach would be "similar" as running wp-now directly from inside the `wp-content` folder.

A more elaborated approach would be mounting each subfolder/file independently resulting on a "merge" between core and the plugin, maybe even recursively if we want that `themes` folder contains twentytwentythree.


This is what I see with the simple approach:
### Themes:
Screenshot 2023-06-02 at 19 22 32


### Plugins:
Screenshot 2023-06-02 at 19 22 47

sejas: I already pushed the simple fix: https://github.com/WordPress/playground-tools/pull/51/commits/1b2260fb6b5ded4f65fd66ece874f62ed7efb35a

danielbachhuber: @sejas Thanks for taking a look at this!

I think I'm leaning towards _not_ shipping this pull request, and instead solving this with https://github.com/WordPress/playground-tools/issues/19

I'm not strongly opinionated about it, though. What do you think?

sejas: Mapping folders is a more general fix, let's go for it. 👍

Issue #33, opened 2023-05-09
Local Environment: Child theme can't be activated


danielbachhuber:
If you run `wp-now --path=/path/to/child-theme`, the theme can't be activated because the parent theme is missing.

The parent theme is missing because it isn't mounted into the filesystem.

**Done is:**
* We're processing the `style.css` file for the parent them.
* If there is a parent theme, we mount the corresponding directory to the filesystem.

sejas:
@katinthehatsite, Do you want to take this one?

katinthehatsite:
let's do this - or at least attempt 😅

danielbachhuber:
@katinthehatsite I'm going to pick this up so you can focus on your upcoming project.

danielbachhuber:
With some more reflection, I actually want to hold off on fixing this for now:
* https://github.com/WordPress/playground-tools/issues/19 might be a more durable fix, because the parent theme can be located in other theme directories.
* It might be nice to automatically install the parent theme from WordPress.org, but I don't know if this is a safe assumption to make.

I'd love to hear more reports from folks who run into this in the wild, and what their scenario is.

cpapazoglou:
> * It might be nice to automatically install the parent theme from WordPress.org, but I don't know if this is a safe assumption to make.

If it doesn't exist locally and if it's a .org theme, yes, it would be nice!

dmsnell:
Noting here that WordPress also fails if the parent theme is missing. If we can mirror WordPress behavior that seems sufficient. If WordPress can auto-install then that seems reasonable, but for many themes we know this is impossible because they don't exist in the registry.

For those cases it seems sufficient to scope-limit this tool to _not_ solve problems WordPress itself doesn't solve.

danielbachhuber:
I'm also not sure if we can know definitively that a given theme is a WordPress.org theme.

cpapazoglou:
> I'm also not sure if we can know definitively that a given theme is a WordPress.org theme.

3pd themes _usually_ don't have a slug used in .org. For example

`https://api.wordpress.org/themes/info/1.1/?action=theme_information&request%5Bslug%5D=astra` returns data for `astra` theme.

wojtekn:
> For those cases it seems sufficient to scope-limit this tool to not solve problems WordPress itself doesn't solve.

I agree with @dmsnell , local environment tool shouldn't solve WordPress problems that way.

[...]

PR #58, opened 2023-06-02
wp-now: Mount the parent theme when running in a child theme

kozer:
## What?

This pr introduces a fix when the user works in a child theme. In this case, the parent theme is not mounted and this creates a "Missing parent theme" error:
child_theme_error

## Why?

This pr is the first attempt to fix that error.
There is plenty of discussion about this problem [here](https://github.com/WordPress/playground-tools/issues/33)

## How?

As a first step to solve this issue, we try to find the parent theme and mount it in the appropriate folder.
We assume that the parent theme is available in the parent directory, and there is some upcoming work to handle more complex scenarios

## Testing Instructions

1. Create a folder with a theme (eg: test), and a child theme (eg: test-child)
2. Run `wp-now start --path=<your-child-theme-path>`
3. Ensure there is no error in the loaded page
4. Move to `/wp-admin/themes.php` and ensure that you:
     - Can activate and deactivate your child theme
     - The parent theme is visible and can be activated and deactivated as well

### DIFF
--- a/packages/wp-now/src/wp-now.ts
+++ b/packages/wp-now/src/wp-now.ts
@@ -241,10 +241,17 @@ async function runPluginOrThemeMode(
 		const templateName = getThemeTemplate(projectPath);
 		if (templateName) {
 			// We assume that the theme template is in the parent directory
-			php.mount(
-				path.join(projectPath, '..', templateName),
-				`${documentRoot}/wp-content/${directoryName}/${templateName}`
-			);
+			const templatePath = path.join(projectPath, '..', templateName);
+			if (fs.existsSync(templatePath)) {
+				php.mount(
+					templatePath,
+					`${documentRoot}/wp-content/${directoryName}/${templateName}`
+				);
+			} else {
+				output?.error(
+					`Parent for child theme not found: ${templateName}`
+				);
+			}
 		}
 	}
 	mountSqlitePlugin(php, documentRoot);
--- a/packages/wp-now/src/wp-now.ts
+++ b/packages/wp-now/src/wp-now.ts
@@ -22,6 +22,7 @@ import {
 	isWordPressDirectory,
 	isWordPressDevelopDirectory,
 	getPluginFile,
+    readFileHead,
 } from './wp-playground-wordpress';
 import { output } from './output';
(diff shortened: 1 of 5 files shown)


packages/wp-now/src/wp-now.ts
```diff
@@ -236,6 +237,16 @@ async function runPluginOrThemeMode(
 		projectPath,
 		`${documentRoot}/wp-content/${directoryName}/${pluginName}`
 	);
+	if (mode === WPNowMode.THEME) {
+		const templateName = getThemeTemplate(projectPath);
+		if (templateName) {
```
danielbachhuber, line 242: Should we make sure the directory exists before mounting?

packages/wp-now/src/wp-now.ts
```diff
@@ -318,6 +329,15 @@ async function activatePluginOrTheme(
 	}
 }
 
+function getThemeTemplate(projectPath: string) {
```
danielbachhuber: Can you add some tests for this new function?

packages/wp-now/src/wp-now.ts
```diff
@@ -236,6 +237,16 @@ async function runPluginOrThemeMode(
 		projectPath,
 		`${documentRoot}/wp-content/${directoryName}/${pluginName}`
 	);
+	if (mode === WPNowMode.THEME) {
```
danielbachhuber, line 240: Can you add some new tests for this behavior?

packages/wp-now/src/wp-now.ts
adamziel: Mount() does that now

danielbachhuber (approved):
Works as expected with one tweak: a479df31b58babe7142adabcc7bb7a4333017f79

I downloaded https://wordpress.org/themes/moina-blog/, unzipped it, and ran `nx preview wp-now start --path=/Users/danielbachhuber/projects/playground-test/child-theme/moina-blog`. It was marked as a broken theme because it was missing its parent:

image

When I downloaded the parent theme (https://wordpress.org/themes/moina/) and unzipped it in the same directory, the child theme activated as expected:

image

PR #59, opened 2023-06-02
Compress text files

luisherranz:
## What?

Add compression to the Express server.

co-authored with @adamziel.

## Why?

Even when developing, it's important to be able to inspect the compressed size of the assets.

Other tools like `@wordpress/env` or Local also do it by default.

## How?

Add the [compression](https://www.npmjs.com/package/compression) middleware.

We had to add our own custom `filter` because the default one was not working with the `content-type` headers that WordPress returns. We added a very basic one, but it should be enough for development.

_The default one uses [this database of types](https://github.com/jshttp/mime-db/blob/master/db.json), but the [regular expressions](https://github.com/jshttp/compressible/blob/master/index.js) were failing._

## Testing Instructions

- Start the server.
- Check that the text files are compressed.

---

I couldn't get the tests to work. For some reason, the `fetch` is failing. I would appreciate some help with that.


adamziel:
cc @sejas @danielbachhuber

danielbachhuber:
Looks cool to me!

luisherranz:
So, after debugging a bit more I've discovered that for some reason [`compression` was not expecting that `res.getHeader` can return an array](https://github.com/expressjs/compression/blob/master/index.js#L254-L263), which in this server configuration does. I've changed that to keep it closer to the original implementation.

I still need to figure out why the `fetch` of the tests is failing, so I appreciate any help in that regard 🙂

packages/wp-now/src/tests/wp-now.spec.ts
```diff
@@ -514,6 +514,42 @@ describe('Test starting different modes', () => {
 		expect(themeName.text).toContain('Twenty Twenty-Three');
 	});
 
+	/**
+	 * Test that startWPNow compresses the text files.
+	 */
+	test('startWPNow compresses the text files', async () => {
+		const projectPath = path.join(tmpExampleDirectory, 'theme-with-assets');
+		const config = await getWpNowConfig({ path: projectPath });
+		const { options } = await startWPNow(config);
```
sejas: After debugging the tests, I found the origin of the error.
`startWPNow` doesn't start the express server, it only prepares the `php-wasm` instance.
We should call to `startServer` to actually start the express server and where your middleware was implemented.

```suggestion
		const { options } = await startServer(config);
```

packages/wp-now/src/tests/wp-now.spec.ts
```diff
@@ -514,6 +514,42 @@ describe('Test starting different modes', () => {
...
+
+		const style = await fetch(
+			`${options.absoluteUrl}/wp-content/themes/theme-with-assets/style.css`
+		);
+		expect(style.headers.get('content-encoding')).toBe('gzip');
+
+		const javascript = await fetch(
+			`${options.absoluteUrl}/wp-content/themes/theme-with-assets/javascript.js`
+		);
+		expect(javascript.headers.get('content-encoding')).toBe('gzip');
```
sejas: The only assertion that is working for me is Style.

Here is the header for html in case it helps:

```
HeadersList {
  [Symbol(headers map)]: Map(5) {
    'x-powered-by' => 'Express',
    'date' => 'Fri, 02 Jun 2023 14:57:01 GMT',
    'connection' => 'keep-alive',
    'keep-alive' => 'timeout=5',
    'transfer-encoding' => 'chunked'
  },
  [Symbol(headers map sorted)]: null
}
```

packages/wp-now/src/tests/wp-now.spec.ts
```diff
@@ -514,6 +514,42 @@ describe('Test starting different modes', () => {
...
+	});
+
+	/**
+	 * Test that startWPNow doesn't compress non text files.
+	 */
+	test("startWPNow doesn't compress non text files", async () => {
+		const projectPath = path.join(tmpExampleDirectory, 'theme-with-assets');
+		const config = await getWpNowConfig({ path: projectPath });
+		const { options } = await startWPNow(config);
```
sejas:
```suggestion
		const { options } = await startServer(config);
```

sejas (commented):
@luisherranz , thanks for creating this PR.

I found why the fetch was failing.

packages/wp-now/src/tests/wp-now.spec.ts
adamziel: @sejas maybe using a clearer distinction between the two functions would reduce the future confusion? startWPNow sounds like it could run the server. Not sure what a better name would be but just wanted to highlight this.

luisherranz:
@sejas, I've changed `startWPNow` with `startServer` and this should be working now.

I had to introduce a new `stopServer` function to stop the server between tests. Feel free to rearrange or replace any code.

sejas (approved):
@luisherranz , Thanks for contributing to `wp-now` and the playground.

 I tested it locally, and it works great! 🥳

PR #60, opened 2023-06-02
wp-now: add wp-admin redirection with trailing slash

sejas:
## What?

It adds a redirection to `/wp-admin` to add the trailing slash `/wp-admin/` so core loads the correct links in the sidebar.

## Why?

- Reported here https://github.com/WordPress/wordpress-playground/issues/461

## How?

I created a middleware to add trailing slash given a path.

## Testing Instructions

- Run `npx nx build wp-now`
- `node dist/packages/wp-now/main.js start`
- Go to `http://127.0.0.1:8881/wp-admin` (without the last slash)
- Observe you are redirected to `http://127.0.0.1:8881/wp-admin/` and the sidebar menu has the correct links.

### DIFF
--- a/packages/wp-now/src/start-server.ts
+++ b/packages/wp-now/src/start-server.ts
@@ -7,6 +7,7 @@ import { portFinder } from './port-finder';
 import { NodePHP } from '@php-wasm/node';
 import startWPNow from './wp-now';
 import { output } from './output';
+import { addTrailingSlash } from './add-trailing-slash';
 
 function requestBodyToMultipartFormData(json, boundary) {
 	let multipartData = '';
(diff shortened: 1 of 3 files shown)


packages/wp-now/src/start-server.ts
```diff
@@ -33,6 +33,16 @@ const requestBodyToString = async (req) =>
 		});
 	});
 
+function addTrailingSlash(path) {
+	return (req, res, next) => {
+		if (req.url === path) {
```
adamziel: Will that work if the URL is `/wp-admin?a=b`?

danielbachhuber: I tested and it does not.

Maybe `addTrailingSlash` should be a standalone function with tests, and then we should have some other function for this middleware.

sejas: - Now it supports get parameters `/wp-admin?a=b` 
- I moved the function to a new file, and included tests.

danielbachhuber (approved):
Seems to work well! Thanks for adding tests 😊

PR #69, opened 2023-06-06
Copy CI tooling from WordPress Playground

danielbachhuber:
## What?

Copies CI tooling improvements from WordPress Playground, specifically:
* https://github.com/WordPress/wordpress-playground/pull/494
* https://github.com/WordPress/wordpress-playground/pull/496

## Why?

* We should keep CI as consistent as possible between the two repositories.
* It's nice to have linting and testing split into separate jobs, so it's easier to see what fails.
* We'll benefit from a slight performance boost.

## Testing Instructions

1. Verify that all tests are run and pass as expected.

### DIFF
--- a/.github/actions/prepare-playground/action.yml
+++ b/.github/actions/prepare-playground/action.yml
@@ -1,0 +1,32 @@
+name: Clone the repository, install dependencies, and configures nx variables
+
+runs:
+    using: 'composite'
+    steps:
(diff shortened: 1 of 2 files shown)


kozer (approved):

kozer:
Tests are passing, as expected! Great work @danielbachhuber !

Issue #11, opened 2023-05-24
Can't start wp-now on Windows OS because of file path fatal


t-hamano:
I have tried wp-now on both Windows host OS and WSL2 and could not start it. Ths may be a problem with my environment, is there anything I should try?

### Commands executed

```
git clone https://github.com/WordPress/twentytwentythree.git
npm install -g @wp-now/wp-now
cd ./twentytwentythree
wp-now start
```

### Occurred Error

```
ReferenceError: Event is not defined
    at Object.<anonymous> (/home/{username}/.volta/tools/image/packages/@wp-now/wp-now/lib/node_modules/@wp-now/wp-now/node_modules/@php-wasm/node/index.cjs:66566:33)
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
    at Module.load (internal/modules/cjs/loader.js:979:32)
    at Function.Module._load (internal/modules/cjs/loader.js:819:12)
    at ModuleWrap.<anonymous> (internal/modules/esm/translators.js:203:29)
    at ModuleJob.run (internal/modules/esm/module_job.js:183:25)
    at async Loader.import (internal/modules/esm/loader.js:178:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
    at async handleMainPromise (internal/modules/run_main.js:59:12)
```

### Environment

Note: Both environments use Volta as their version control tool.

#### Windows Host

- OS: Windows OS 11
- Node Version: 14.21.1

### WSL2

- OS: Ubuntu 20.04.4 LTS
- Node Version: Tested with both 14.21.3 and 16.18.0

adamziel:
@t-hamano You'll need Node.js 18+, the error message isn't very clear about that unfortunately – I'd love to clarify it.

t-hamano:
@adamziel 
Thanks for the info!

I was able to move beyond the error in Windows OS by changing the Node version to 18, but another error occurred.

<details><summary>Error Log</summary>

```
wp-now start
Starting the server......
directory: D:\Desktop\twentytwentythree
mode: theme
php: 8.0
wp: latest
Downloading WordPress latest...
Downloading SQLite...
Error: Could not mount C:\Users\tetsuaki.hamano\.wp-now\mu-plugins: There is no such file or directory OR the parent directory does not exist.
[...]

PR #71, opened 2023-06-06
Fix fatal startup crash on windows caused by an invalid mount() path

adamziel:
## Description

Solves #11

`wp-now` mounts mu-plugins like this:

```ts
    php.mount(
        path.join(getWpNowPath(), 'mu-plugins'),
        path.join(vfsDocumentRoot, 'wp-content', 'mu-plugins')
    );
```

The first path is the local HOST path, the second path is a VFS path.

On Mac and linux, they both use forward slashes.

On Windows, the first path can use backslashes, but the VFS path still needs to use forward slashes.

Problem is: `path.join()` transforms all slashes to backslashes on Windows, which makes the second path invalid like `\\var\\www\\html`.

This commit corrects that.

## Testing Instructions

Confirm wp-now and the VS code extension can both be started on Windows. Confirm they didn't break on Mac or linux.

### DIFF
--- a/packages/wp-now/src/wp-now.ts
+++ b/packages/wp-now/src/wp-now.ts
@@ -353,7 +353,7 @@ function mountSqliteDatabaseDirectory(
 	fs.ensureDirSync(path.join(wpContentPath, 'database'));
 	php.mount(
 		path.join(wpContentPath, 'database'),
-		path.join(vfsDocumentRoot, 'wp-content', 'database')
+		`${vfsDocumentRoot}/wp-content/database`
 	);
 }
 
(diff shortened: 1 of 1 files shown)


sejas (approved):
@adamziel , Thanks for finding out the root of Windows issues.

Should we update the code related to mounting the database too?
https://github.com/WordPress/playground-tools/blob/4a0ea12abccca970aaf8e233189351b465f1a153/packages/wp-now/src/wp-now.ts#L352-L355

sejas:
> @adamziel , Thanks for finding out the root of Windows issues.
> 
> Should we update the code related to mounting the database too?
> 
> https://github.com/WordPress/playground-tools/blob/4a0ea12abccca970aaf8e233189351b465f1a153/packages/wp-now/src/wp-now.ts#L352-L355

I think `path.join` was working correctly when the first string (`vfsDocumentRoot`) had already `/`. But for consistency, I've removed `path.join` for the database path too.
https://github.com/WordPress/playground-tools/pull/71/commits/614dc1d3bd318d6940ba256923f000d8557a5ec2

danielbachhuber (approved):
Nice find, @adamziel ! 👏

PR #75, opened 2023-06-08
wp-now: Fix php cli path error in dev mode

### LINKED ISSUE #16: wp-now: php command doesn't work for relative path
When I use wp-now in development mode, symlinked using `npm link`, and I try executing PHP script in the current directory, I get an error:

```
$ wp-now php test.php 
[...nx build output here...]
Error: Could not open input file: /Volumes/Sites/wordpress-playground/test.php
    at executePHPFile (file:///Volumes/Sites/wordpress-playground/dist/packages/wp-now/main.js:797:11)
    at async Object.handler (file:///Volumes/Sites/wordpress-playground/dist/packages/wp-now/main.js:895:9)
```

When I use an absolute path, it works:

```
$ wp-now php /Volumes/Sites/vip-directory/test.php
[...nx build output here...]
Hello World!
```

When I install it using npm, it works in both cases:

```
$ wp-now php test.php                                     
Hello World!
```

**Done is**
- I can use relative path parameter in both development and compiled modes

wojtekn:
Fixes https://github.com/WordPress/playground-tools/issues/16

## What?

I get an error when I use wp-now in development mode, symlinked using the `npm link`, and try executing the PHP script in the current directory.

## Why?

`wp-now php` should work in both cases consistently.

## How?

PR adds a code that replaces the second argument with its absolute version if the absolute version exists in the filesystem.

## Testing Instructions

1. Create a hello world PHP script
2. Run it with an absolute path
```
$ wp-now php /Volumes/Sites/vip-directory/test.php
```
3. Run it using the relative path
```
$ wp-now php test.php
```

Test both cases for wp-now installed using npm and for wp-now symlinked using `npm link`.

### DIFF
--- a/packages/wp-now/src/execute-php.ts
+++ b/packages/wp-now/src/execute-php.ts
@@ -31,9 +31,11 @@ export async function executePHP(
 
 	try {
 		php.useHostFilesystem();
-		const maybePhpFile = path.join(wpNowOptions.projectPath, phpArgs[1]);
-		if (fs.existsSync(maybePhpFile)) {
-			phpArgs[1] = maybePhpFile;
+		if (!path.isAbsolute(phpArgs[1])) {
+			const maybePhpFile = path.join(wpNowOptions.projectPath, phpArgs[1]);
+			if (fs.existsSync(maybePhpFile)) {
+				phpArgs[1] = maybePhpFile;
+			}
 		}
 		await php.cli(phpArgs);
 	} catch (resultOrError) {
--- a/packages/wp-now/src/execute-php.ts
+++ b/packages/wp-now/src/execute-php.ts
@@ -30,7 +30,10 @@ export async function executePHP(
 
 	try {
 		php.useHostFilesystem();
-		phpArgs[1] = path.join(wpNowOptions.projectPath, path.basename(phpArgs[1]));
+		phpArgs[1] = path.join(
+			wpNowOptions.projectPath,
+			path.basename(phpArgs[1])
+		);
 		await php.cli(phpArgs);
 	} catch (resultOrError) {
 		const success =
--- a/packages/wp-now/src/execute-php.ts
+++ b/packages/wp-now/src/execute-php.ts
@@ -32,7 +32,10 @@ export async function executePHP(
 	try {
 		php.useHostFilesystem();
 		if (!path.isAbsolute(phpArgs[1])) {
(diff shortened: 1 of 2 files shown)


packages/wp-now/src/execute-php.ts
```diff
@@ -21,14 +23,21 @@ export async function executePHP(
...
-	const { phpInstances } = await startWPNow({
+	const { phpInstances, options: wpNowOptions } = await startWPNow({
 		...options,
 		numberOfPhpInstances: 2,
 	});
 	const [, php] = phpInstances;
 
 	try {
 		php.useHostFilesystem();
+		const maybePhpFile = path.join(
```
wojtekn: It fixes the issue, but I don't like that implementation. Any thoughts on how to make it better?

danielbachhuber: Yeah, I don't love it either.

I think the user's expectation is that the provided file path is relative to their current working directory. Can we prepend the current working directory if the file path isn't absolute?

wojtekn: @danielbachhuber I've added a check for the absolute path, but we still need to check if the path exists as with the current implementation. The reason is that this parameter can be `-v` for the `wp-now php -v` call or `--path=/Sites/my-site` for the `wp-now php` call. 

I've added https://github.com/WordPress/playground-tools/issues/76 to improve php command parameters usage.

kozer:
I tested it @wojtekn! Nice work!

### Before

- **Linked. Relative path:** Throws an error, script is not executed
- **Linked: Absolute path:** Script executed successfully
- **Installed: Relative path:** Script executed successfully
- **Installed: Absolute path:** Script executed successfully

### After

- **Linked. Relative path:** Script executed successfully
- **Linked: Absolute path:** Script executed successfully
- **Installed: Relative path:** Script executed successfully
- **Installed: Absolute path:** Script executed successfully

kozer (approved):

sejas:
I wonder if we can fix the current `npm link` instead.

wojtekn:
@sejas I'm not sure, note that we reproduce it even without an `npm link`, and I've added a test case for that:

https://github.com/WordPress/playground-tools/pull/75/files#diff-cf2575f8a7b2d773ad35151a3979fb1c79a01e3394e6d6d3a25e9a2328123bf9

adamziel (approved):

Issue #62, opened 2023-06-02
Submit the Interactive Code Block to the WordPress Plugin directory


adamziel:
The [Interactive Code Block](https://github.com/WordPress/playground-tools/tree/trunk/packages/interactive-code-block) merged in #168. The initial release works and could already help developers write tutorials if it was available in the plugin directory. Let's submit it there!

adamziel:
<img width="950" alt="CleanShot 2023-06-09 at 14 53 08@2x" src="https://github.com/WordPress/playground-tools/assets/205419/c007c2fd-9433-43f2-a3f5-a762dbf13dd8">

Let's wait for the review!

adamziel:
https://wordpress.org/plugins/interactive-code-block/

PR #77, opened 2023-06-09
Prepare Interactive Code Block for plugin directory submission

adamziel:
## What?

Needed for https://github.com/WordPress/playground-tools/issues/62

* Adds the README file
* Adds the LICENSE file
* Bundles both of them
* Restricts the supported PHP version to 7.4 to keep the bundle size below 10MB (WP.org limit)

## Testing Instructions

Submit a plugin to the directory...? 😆

### DIFF
--- a/packages/interactive-code-block/project.json
+++ b/packages/interactive-code-block/project.json
@@ -13,9 +13,17 @@
 			"options": {
 				"cwd": "dist/packages",
 				"commands": [
+					"cp ../../packages/interactive-code-block/README.md ./interactive-code-block/",
+					"rm interactive-code-block/assets/php_5*",
+					"rm interactive-code-block/assets/php_7_0*",
+					"rm interactive-code-block/assets/php_7_1*",
+					"rm interactive-code-block/assets/php_7_2*",
+					"rm interactive-code-block/assets/php_7_3*",
+					"rm interactive-code-block/assets/php_8*",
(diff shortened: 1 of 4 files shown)


packages/interactive-code-block/project.json
```diff
@@ -13,9 +13,17 @@
 			"options": {
 				"cwd": "dist/packages",
 				"commands": [
+					"cp ../../packages/interactive-code-block/README.md ./interactive-code-block/",
```
adamziel, line 16: I don't love this but I don't have any better idea for now

packages/interactive-code-block/src/lib/php-worker.ts
```diff
@@ -6,18 +6,9 @@ import {
...
-} from '@php-wasm/universal';
+import { SupportedPHPVersionsList } from '@php-wasm/universal';
 
-interface WorkerOptions extends Record<string, string> {
-	php: string;
-}
-const options = parseWorkerStartupOptions<WorkerOptions>();
-const phpVersion = (options?.php?.replace('_', '.') ||
-	LatestSupportedPHPVersion) as SupportedPHPVersion;
+const phpVersion = '7.4';
```
adamziel, line 11: Just hardcodes PHP 7.4, no point in supporting other versions if they're not getting bundled

Issue #9, opened 2023-05-23
Local Environment: Replace custom setup logic with a Blueprints-based setup


adamziel:
wp-now implements:

* Downloading and extracting zip files
* Download parallelization
* Using the requested PHP / WordPress version via CLI options
* Aims to have a JSON config at some point

These are Blueprints features. There's no need to re-implement them from scratch. Let's use and contribute to a common standard that other Playground projects will immediately benefit from.

For example, [wp-now: download wordpress and sqlite in parallel](https://github.com/WordPress/wordpress-playground/pull/417) wouldn't be needed if the assets were downloaded through a Blueprint.

cc @sejas @danielbachhuber @wojtekn @katinthehatsite

danielbachhuber:
How do Blueprints work?

adamziel:
I wrote up this documentation last week: https://wordpress.github.io/wordpress-playground/docs/blueprints-api/index

In short, it's a declarative data format where you communicate all the steps needed to set up a WordPress instance, and then the executor finds the optimal way of running these steps. They work in the browser and in Node.js.

For now this means downloading files in parallel with a concurrency limit. In the future it will involve flattening and parallelizing entire steps. 

This Blueprint, for example, downloads a plugin and a theme in parallel and then installs both:

```json
{
    "steps": [
        {
            "step": "installPlugin",
            "pluginZipFile": {
                "resource": "wordpress.org/plugins",
                "slug": "coblocks"
            }
        },
        {
            "step": "installTheme",
            "pluginZipFile": {
                "resource": "wordpress.org/themes",
                "slug": "pendant"
            }
        }
    ]
}
```

This one resolves the PHP and WordPress version to use:

```json
{
    "preferredVersions": {
        "php": "7.4",
        "wp": "5.9"
    }
}
```

^ Blueprints will be merge-able, which means I'll be able to take a Blog Blueprint and a Woo store Blueprint and create a Woo Blog that uses the right PHP and WordPress versions.

adamziel:
An update on that last point: https://github.com/WordPress/blueprints/issues/49

JUVOJustin:
I love the concept and do think blueprints are an awesome approach. For my specific use case of using the playground as a wp-env replacement, it would be great to have a way to pass a path to a blueprint for my local project as an optional parameter to wp-now. Maybe something like `wp-now start -b './blueprint.json'`.  Some more specific thoughts:

- Ideally it would automatically use something like playground.blueprint.json by default and as fallback
[...]

PR #78, opened 2023-06-14
wp-now: Add support for custom URL's

sergeymitr:
## What?

Support for custom URL's and tunneling services (e.g. ngrok).

## Why?

Useful for development, requested by users.

## How?

We add two new CLI options:
- `--url` replaces `WP_HOME` and `WP_SITEURL` values for the site to support custom domain names
- `--customport` is needed for proxying tunneling services (e.g. ngrok)

For local custom domains, it's enough to add the site record into the `hosts` file:
```
127.0.0.1 wpnow-test.local
```
And then pass that value as `--url="http://wpnow-test.local"`
The website will be available at `http://wpnow-test.local:[port]`, with `[port]` being the open port used by wp-now in regular fashion.
The final URL will be displayed in shell and loaded automatically in browser instead of the localhost one:
```
Custom site URL: http://wpnow-test.local:8881
```

If you use ngrok or similar tunneling service, you will need to pass both `--url="http://your-ngrok-domain"` and `--customport=80` because proxying to 8881 (or whatever is generated by wp-now) happens automatically.

Also, there's still no HTTPS supported, so make sure you use `http://` for whatever domain you're proxying.

## Testing Instructions

0. Checkout the branch.

### Local custom domain 
1. Add your custom domain into `hosts` file (e.g. `127.0.0.1 wpnow.local`)
2. Run wp-now with the argument `--url="http://wpnow.local"
3. That URL should get opened in the browser, the site should work fine.

### Tunneling service
1. Establish ngrok proxying: `ngrok http localhost:8882 --scheme=http` (no HTTPS supported yet)
2. Run wp-now: `nx preview wp-now start --port=8882 --url="http://your-ngrok-domain.something" --customport=80`
3. The ngrok domain should load automatically in the browser.
4. If the browser keeps redirecting you to https:// and you aren't able to load the site, try good ol' `curl -i http://your-ngrok-domain.something`

### DIFF
--- a/packages/wp-now/src/run-cli.ts
+++ b/packages/wp-now/src/run-cli.ts
@@ -39,7 +39,7 @@ function commonParameters(yargs) {
 		})
 		.option('customport', {
 			describe:
-				"Custom port number. Needed if you're using a reverse proxy (e.g. ngrok), usually 80.",
+				"Custom port number. Needed if you're using a tunneling service (e.g. ngrok), usually 80.",
 			type: 'number',
 		});
 }
--- a/packages/wp-now/src/run-cli.ts
+++ b/packages/wp-now/src/run-cli.ts
@@ -32,7 +32,7 @@ function commonParameters(yargs) {
 			describe: 'PHP version to use.',
 			type: 'string',
 		})
-		.option('siteurl', {
+		.option('url', {
 			describe:
 				"Custom site URL to access the site ('home' and 'siteurl' option values).",
 			type: 'string',
@@ -86,7 +86,7 @@ export async function runCli() {
 						php: argv.php as SupportedPHPVersion,
 						wp: argv.wp as string,
 						port: argv.port as number,
-						siteurl: argv.siteurl as string,
+						url: argv.url as string,
 						customport: argv.customport as number,
 					});
 					portFinder.setPort(options.port as number);
--- a/packages/wp-now/src/run-cli.ts
+++ b/packages/wp-now/src/run-cli.ts
@@ -31,6 +31,16 @@ function commonParameters(yargs) {
 		.option('php', {
 			describe: 'PHP version to use.',
 			type: 'string',
+		})
+		.option('siteurl', {
+			describe:
+				"Custom site URL to access the site ('home' and 'siteurl' option values).",
+			type: 'string',
+		})
+		.option('customport', {
+			describe:
+				"Custom port number. Needed if you're using a reverse proxy (e.g. ngrok), usually 80.",
+			type: 'number',
 		});
 }
 
@@ -76,6 +86,8 @@ export async function runCli() {
 						php: argv.php as SupportedPHPVersion,
 						wp: argv.wp as string,
 						port: argv.port as number,
+						siteurl: argv.siteurl as string,
+						customport: argv.customport as number,
 					});
 					portFinder.setPort(options.port as number);
 					const { url } = await startServer(options);
--- a/packages/wp-now/src/config.ts
+++ b/packages/wp-now/src/config.ts
@@ -18,6 +18,8 @@ export interface CliOptions {
 	path?: string;
 	wp?: string;
 	port?: number;
+	siteurl?: string;
+	customport?: number;
 }
 
 export const enum WPNowMode {
@@ -41,6 +43,8 @@ export interface WPNowOptions {
 	wpContentPath?: string;
 	wordPressVersion?: string;
 	numberOfPhpInstances?: number;
+	customSiteURL?: string;
+	customPort?: number;
 }
 
 export const DEFAULT_OPTIONS: WPNowOptions = {
@@ -50,6 +54,8 @@ export const DEFAULT_OPTIONS: WPNowOptions = {
 	projectPath: process.cwd(),
 	mode: WPNowMode.AUTO,
 	numberOfPhpInstances: 1,
+	customSiteURL: null,
+	customPort: null,
 };
 
 export interface WPEnvOptions {
@@ -96,6 +102,8 @@ export default async function getWpNowConfig(
 		projectPath: args.path as string,
 		wordPressVersion: args.wp as string,
 		port,
+		customSiteURL: args.siteurl as string,
+		customPort: args.customport as number,
 	};
 
 	const options: WPNowOptions = {} as WPNowOptions;
@@ -120,6 +128,15 @@ export default async function getWpNowConfig(
 	if (!options.absoluteUrl) {
 		options.absoluteUrl = await getAbsoluteURL();
 	}
+
+	if (options.customSiteURL) {
+		const customPort =
+			options.customPort || (await portFinder.getOpenPort());
+		if (customPort && customPort !== 80) {
+			options.customSiteURL += `:${customPort}`;
+		}
+	}
+
 	if (!isValidWordPressVersion(options.wordPressVersion)) {
 		throw new Error(
 			'Unrecognized WordPress version. Please use "latest" or numeric versions such as "6.2", "6.0.1", "6.2-beta1", or "6.2-RC1"'
--- a/packages/wp-now/src/config.ts
+++ b/packages/wp-now/src/config.ts
@@ -18,7 +18,7 @@ export interface CliOptions {
 	path?: string;
 	wp?: string;
 	port?: number;
-	siteurl?: string;
+	url?: string;
 	customport?: number;
 }
 
@@ -102,7 +102,7 @@ export default async function getWpNowConfig(
 		projectPath: args.path as string,
 		wordPressVersion: args.wp as string,
 		port,
-		customSiteURL: args.siteurl as string,
+		customSiteURL: args.url as string,
 		customPort: args.customport as number,
 	};
 
--- a/packages/wp-now/src/start-server.ts
+++ b/packages/wp-now/src/start-server.ts
@@ -103,10 +103,14 @@ export async function startServer(
 			output?.trace(e);
 		}
 	});
-	const url = options.absoluteUrl;
+	const internalUrl = options.absoluteUrl;
+	const customSiteUrl = options.customSiteURL;
 	app.listen(port, () => {
-		output?.log(`Server running at ${url}`);
+		output?.log(`Server running at ${internalUrl}`);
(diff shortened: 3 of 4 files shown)


packages/wp-now/src/run-cli.ts
```diff
@@ -31,6 +31,16 @@ function commonParameters(yargs) {
...
+		})
+		.option('siteurl', {
+			describe:
+				"Custom site URL to access the site ('home' and 'siteurl' option values).",
+			type: 'string',
+		})
+		.option('customport', {
+			describe:
+				"Custom port number. Needed if you're using a tunneling service (e.g. ngrok), usually 80.",
+			type: 'number',
```
sejas: I wonder if we can/should remove `customport` and consider it always part of `siteurl`.
Assuming that if a given `siteurl` is passed the custom port will be `80` and if it's a different port, it needs to be explicit on `siteurl` string.

sejas (commented):
@sergeymitr  Thanks for the PR!, I'll try it later.
The code looks good so far, I just wonder if `customport` is necessary.

Another approach to change the `siteurl` could be accepting blueprints.

packages/wp-now/src/run-cli.ts
danielbachhuber: I'd prefer this too.

packages/wp-now/src/run-cli.ts
```diff
@@ -31,6 +31,16 @@ function commonParameters(yargs) {
 		.option('php', {
 			describe: 'PHP version to use.',
 			type: 'string',
+		})
+		.option('siteurl', {
```
danielbachhuber: Can we call this argument `url`?

danielbachhuber (commented):
Rather than introducing a new argument, I wonder if we'd be better off incorporating Blueprint support (https://github.com/WordPress/playground-tools/issues/9) and then using the Blueprint to set these values.

packages/wp-now/src/run-cli.ts
sergeymitr: Thanks for taking a look!
The problem with `customport` is that local custom domains (the `hosts` file ones) don't do port forwarding, so we'll need to add the port explicitly to the custom domain, e.g. `http://somethingcool.local:8881`. Since the port can differ depending on whichever's open, it cannot be passed as an option.

Tunneling services, on the other hand, do port forwarding, so `http://somethingcool.ngrok.something` (port 80) will forward to `http://localhost:8881`. So the port should be in the URL for local custom domains, and shouldn't be there for non-custom domains.

At the very least we would need a flag, something like `--addPortNumberToTheSiteURL` to determine whether to add that wp-now-determined port number to the site URL or not.
I can imagine the situation when tunneling service port differs from 80, that's why I made it a port number and not a flag.

I didn't manage to come up with a way to get rid of it altogether, so if I'm missing something, please let me know.

sergeymitr: After thinking about it, I guess we could take that port number from the URL, and get rid of the `customport` option.
That would mean 3 possible states:
- port 80: no port in the final URL for tunneling service
- non-80 port specified: the final URL will contain that port for tunneling service with non-80 port for some reason
- no port: port will be automatically added to the URL by wp-now

The whole situation is a bit user-unfriendly though, both with `customport` option and without it.

May be we should replace it with a flag (e.g. `--proxying`) to indicate that the final URL does not need to get the port added to it?
That wouldn't work for proxying via non-80 ports, but I don't even know if that's a thing. That potential flexibility may not be worth the added complexity.

What do you think about that?

danielbachhuber: The more that I think about this, the more I feel we should solve this with Blueprints (https://github.com/WordPress/playground-tools/issues/9).

@adamziel Could you provide a sample Blueprint that would achieve what @sergeymitr wants?

sergeymitr: That might be a good solution.
Is there a way to retrieve the [wp-now port](https://github.com/WordPress/playground-tools/blob/trunk/packages/wp-now/src/port-finder.ts#L42) from a blueprint?

packages/wp-now/src/run-cli.ts
sergeymitr: Done!

packages/wp-now/src/run-cli.ts
danielbachhuber: Created a GitHub issue for @sejas to work on the Blueprint integration: https://github.com/WordPress/playground-tools/issues/80

adamziel: @sergeymitr Is this just to define the WordPress constants? If so, this should work:

```json
{
    "steps": [
         { "step": "defineSiteUrlStep", "siteUrl": "http://myurl.localhost" }
     ]
}
```

If you need some extra logic there you could either have three files or generate that blueprint in a custom bash script to startup the dev env.

sergeymitr: Thanks @adamziel.

After additional research, I don't think blueprints would solve the issue entirely:
- port number is being determined dynamically later in the code, so we'd still need to edit the `getWpNowConfig()` function to add it to the URL
- decision whether to add it to the URL (for local domains) or not (for reverse proxies/tunneling services)

Considering changes in the code will still be needed, blueprints don't bring significant improvement. 
Since you're striving toward keeping the tools lean (I fully support this), I suggest one of the approaches below.
I believe they would bring the needed flexibility to solve a lot of edge cases that users will come up with:

1. Do not define the constants `WP_HOME` and `WP_SITEURL`, and save them in the database options table just like regular WordPress installations do.
2. Depending on a flag (e.g. `--extract-url-config`) or some blueprint configuration, add `WP_HOME` and `WP_SITEURL` into the existing `wp-config.php`, or create a new one.
3. Use a different flag (e.g. `--skip-url-config`) or a blueprint to skip setting `WP_HOME` and `WP_SITEURL`. Thus, the site will not work until they are defined in a custom `wp-config.php`.
4. Use another flag (e.g. `--auto-url-config`) or a blueprint to make `WP_HOME` and `WP_SITEURL` take whatever the value is in the environment variables ([example from Jetpack dev env](https://github.com/Automattic/jetpack/blob/ecb8f7d24fa4dde0831f84b0f4d7b6af9a086c72/tools/docker/bin/run.sh#L51-L56)).

It's likely that the issue of hardcoded `WP_HOME` and `WP_SITEURL` will eventually come up, as not being able to change the URL setting of the site limits the ability to develop and test quite a bit. Right now the `wp_option` records get ignored, trying to customize it in `wp-config.php` throws the "Constant already defined" error.

What do you think if we implement one of those approaches, or may be you have other ideas?

I believe being able to edit those values via `wp-config.php` as intended by WordPress would be a better approach than having to implement and apply blueprints to do the same.

sejas: > I believe being able to edit those values via wp-config.php as intended by WordPress would be a better approach than having to implement and apply blueprints to do the same.

@sergeymitr, I totally agree. The more we keep the natural behaviour, the best.

Currently, those config variables are set here.

https://github.com/WordPress/playground-tools/blob/6602715d31e2f23441735910af8ea73d7c08ceb6/packages/wp-now/src/wp-now.ts#L327-L337

It's possible to apply and overwrite those variables with a blueprint step. Which is very similar to your second option:

> Depending on a flag (e.g. --extract-url-config) or some blueprint configuration, add WP_HOME and WP_SITEURL into the existing wp-config.php, or create a new one.

I already created a PR to support the blueprints. I still need to validate your case, and I'm confident that it will fix your use case 🤞.

- https://github.com/WordPress/playground-tools/pull/82

sergeymitr: Hi @sejas.

Speaking of my task at hand, the blueprints would allow using tunneling services, so we're good here.
However, I don't think they'll solve the issue entirely.

WP-now starts a server on a random port (even if you provide port number, it may change if the port is closed). Therefore, custom local URL's will look like this: `http://something.local:8765`. Since we cannot know the port number before starting the server, we'll still need to edit the code to append the port number to the URL provided in the blueprint.

However, tunneling services (ngrok, JT) proxy requests straight to that custom port, making you access the site via port 80. So, the website URL will look like `http://something-ngrok.tld`, with no port number needed.

Blueprints should work for the tunneling services, since there's no need to add port number to the site URL. However, they will not work for regular custom domains because we need to know the port in advance, or append it to the URL. If we go with the latter, we also need to somehow inform the script whether we actually need to append the port.

It also limits the users because of the non-native way to define the site URL. One thing that comes to mind is running multi-URL sites. While `wp-config.php` allows us to simply use environment variables to make the site work on any domain, blueprints don't do that.
They'll also need to restart the server every time the domain changes.

Blueprints are undoubtedly a great way to customize the server, but I still think making `WP_HOME` and `WP_SITEURL` immutable is quite limiting.

sejas: - @sergeymitr I merged https://github.com/WordPress/playground-tools/pull/82 , It's working on Node v20. Feel free to test it. We didn't release the package yet.

> WP-now starts a server on a random port (even if you provide port number, it may change if the port is closed). Therefore, custom local URL's will look like this: http://something.local:8765. Since we cannot know the port number before starting the server, we'll still need to edit the code to append the port number to the URL provided in the blueprint.

The default port is not random, it's usually `8881`. You are right that this port could change if that port is already in use, but it increases sequentially. So it's a deterministic way and users can predict what port will run wp-now.

Since the port is predictable, the user can write the given port manually and include it on WP_SITEURL through a blueprint.
I'm still thinking that passing the `--url` parameter is something that can be handy for wp-now, but this `url` can also include the default port or the same port passed to `--port` e.g. `--url=http://myurl.wp-now:8881` or `--url=http://myurl.wp-now --port=80`

> Blueprints are undoubtedly a great way to customize the server, but I still think making WP_HOME and WP_SITEURL immutable is quite limiting.

I won't be available next week, but we can keep iterating this PR and probably accept `--url` for a fast option change.

sergeymitr: Hi @sejas.
Your blueprint PR works great, thanks for implementing that!
It would come handy if we could edit `WP_HOME` and `WP_SITEURL` right in `wp-config.php`, but that's no longer a priority since everything works with blueprints.
I'm going to close the PR now as the changes no longer necessary, and we can revisit this later if needed.

PR #92, opened 2023-08-01
wp-now: Require node v18 or higher to run

dmsnell:
In this patch we're calling a wrapper script that checks the version of Node before executing code from the project.

Previously, the version check occurred within the same running Node process as the project code, and so on unsupported Node versions it could crash while reading in the initial code to run.

By moving the check out of process we can ensure that the minimum syntax and library support is available before we attempt to run any code.

cc: @danielbachhuber 

I haven't yet determined if I can run this with the wrapper script, but I independently tested the wrapper script and I think it's reliable enough. As for your questions about about command-line arguments I believe it's taken care of by using the `"$@"` in quotes like that; it's a special form added exactly to fix broken string interpolation issues with arguments.


dmsnell:
I'm not sure if Node ever reports anything after the version numbers like `v20.4.15-beta` but if it does we would need to update the pattern in this file. I've tried to be conservative as a first step to avoid making it fuzzy

packages/wp-now/public/with-node-version.sh
```diff
@@ -0,0 +1,28 @@
...
+MIN_PATCH=0
+
+VERSION=( $(node -e "console.log(/^v(\d+)\.(\d+)\.(\d+)$/.exec(process.version).slice(1, 4).join(' '))") )
+
+MAJOR=${VERSION[0]}
+MINOR=${VERSION[1]}
+PATCH=${VERSION[2]}
+
+if [ $MAJOR -lt $MIN_MAJOR ]; then
+	printf "This script is requires node version v$MIN_MAJOR.$MIN_MINOR.$MIN_PATCH or above; found v$MAJOR.$MINOR.$PATCH."
```
danielbachhuber:
```suggestion
	printf "This script is requires node version v$MIN_MAJOR.$MIN_MINOR.$MIN_PATCH or above; found v$MAJOR.$MINOR.$PATCH"
```

I find the periods to add a little of confusion to the output.

packages/wp-now/public/with-node-version.sh
```diff
@@ -0,0 +1,28 @@
...
+	exit 1
+fi
+
+if [ $MINOR -lt $MIN_MINOR ]; then
+	printf "This script is requires node version v$MIN_MAJOR.$MIN_MINOR.$MIN_PATCH or above; found v$MAJOR.$MINOR.$PATCH."
```
danielbachhuber:
```suggestion
	printf "This script is requires node version v$MIN_MAJOR.$MIN_MINOR.$MIN_PATCH or above; found v$MAJOR.$MINOR.$PATCH"
```

Ditto above.

packages/wp-now/public/with-node-version.sh
```diff
@@ -0,0 +1,28 @@
...
+	exit 1
+fi
+
+if [ $PATCH -lt $MIN_PATCH ]; then
+	printf "This script is requires node version v$MIN_MAJOR.$MIN_MINOR.$MIN_PATCH or above; found v$MAJOR.$MINOR.$PATCH."
```
danielbachhuber:
```suggestion
	printf "This script is requires node version v$MIN_MAJOR.$MIN_MINOR.$MIN_PATCH or above; found v$MAJOR.$MINOR.$PATCH"
```

Ditto above.

danielbachhuber (commented):
> I haven't yet determined if I can run this with the wrapper script, but I independently tested the wrapper script and I think it's reliable enough.

@dmsnell How did you test it? I'm getting some errors:

```
$ node /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".sh" for /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh
```

It seems to work with bash, but I'm not sure node is smart enough to call it with bash:

```
$ bash /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start
This script is requires node version v18.0.0 or above; found v14.21.3.
```

dmsnell:
> How did you test it?

sorry, I tried to clarify this in the description - I didn't test this

> `node /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start`

hm. was this previously being invoked with `node`? I thought that we typically invoked `cli.js`, which is why it has the hashbang for `node`, so that the shell runs it with node.

my hope was this would work as a drop-in replacement, but maybe the `bin` value there in `package.json` implies running with node?

danielbachhuber:
> was this previously being invoked with `node`? I thought that we typically invoked `cli.js`, which is why it has the hashbang for `node`, so that the shell runs it with node.

That was my assumption but your assumption seems reasonable too.

Any ideas why I get this error when trying to directly invoke the script?

```
$ ./Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start
zsh: no such file or directory: ./Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh
$ head /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh
#!/usr/bin/env bash

MIN_MAJOR=18
MIN_MINOR=0
MIN_PATCH=0

VERSION=( $(node -e "console.log(/^v(\d+)\.(\d+)\.(\d+)$/.exec(process.version).slice(1, 4).join(' '))") )

MAJOR=${VERSION[0]}
MINOR=${VERSION[1]}
```

dmsnell:
> Any ideas why I get this error when trying to directly invoke the script?

Looks like you're trying to invoke it from the current directory but giving it an absolute address. If you remove the leading `.` before `./Users/…` it should work.

As a side note: it's required to add the dot when running something that isn't in the `PATH` for your shell, because it needs to form a path to the executable. If, however, you provide an absolute address, it doesn't need the dot because it already has the full path.

```zsh
/Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.sh /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start
```

or

```zsh
cd /Users/danielbachhuber/projects/playground-tools
dist/packages/wp-now/with-node-version.sh dist/packages/wp-now/cli.js start
```

or

```zsh
cd ~/projects/playground-tools
./dist/packages/wp-now/with-node-version.sh ./dist/packages/wp-now/cli.js start
```

these both need to be executable, so we may have to make sure that happens automatically when building the `dist` folder. if it tells you it doesn't have permission try calling `chmod +x dist/packages/wp-now/…`

packages/wp-now/public/with-node-version.sh
dmsnell: dots removed in 374b1370

packages/wp-now/public/with-node-version.sh
dmsnell: ditto above.

packages/wp-now/public/with-node-version.sh
dmsnell: ditto above.

dmsnell:
Removed the dots in 374b1370, though I know people around these parts like the trailing periods 🙃 

I will try now to write the wrapper script in JavaScript and pass off execution the same way that the BASH script is doing.

dmsnell:
@danielbachhuber you might try this again. I've put in a `node` wrapper script, hoping it runs now when invoked from `npm`. it seems to work for me when I call it directly, but I still am not entirely sure how to test it before publishing the package.

danielbachhuber (approved):
The Node wrapper script seems to work well.

**`start` positional arg**

```
$ node /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.js /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start
Starting the server......
directory: /Users/danielbachhuber/Desktop
mode: playground
php: 8.0
wp: latest
Server running at http://localhost:8881
```

**`--php=<version>` associative arg**

```
$ node /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.js /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js start --php=8.1
Starting the server......
directory: /Users/danielbachhuber/Desktop
mode: playground
php: 8.1
wp: latest
Server running at http://localhost:8881
```

**Unsupported version**

```
$ node /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/with-node-version.js /Users/danielbachhuber/projects/playground-tools/dist/packages/wp-now/cli.js

This script is requires node version v18.0.0 or above; found v14.21.3
```

I'm going to push a release to test.

PR #97, opened 2023-08-16
Update webview.tsx

lemanschik:
## What?
Added information for debugging if something goes wrong

## Why?
installed wrong plugins and it did not load anymore needed a way to manage the installation

## How?
Add information for the user so he knows in case where to search for more info

## Testing Instructions
No testing needed i guess the text works

### DIFF
--- a/packages/vscode-extension/src/webview.tsx
+++ b/packages/vscode-extension/src/webview.tsx
@@ -59,6 +59,12 @@ const Webview = ({ initialState }: WebviewProps) => {
 				below to start the playground:
 			</p>
 			<ServerState appState={appState} />
+			<p>
+				This Extension is based on https://github.com/WordPress/playground-tools/wp-now
+				so the default storage location is ~/.wp-now if there are any problems you can 
+				fix them via the wp-now cli or if you do not depend on the state even delete 
+				the directory it gets recreated. 
+			</p>
 		</div>
 	);
 };


packages/vscode-extension/src/webview.tsx
```diff
@@ -59,6 +59,12 @@ const Webview = ({ initialState }: WebviewProps) => {
 				below to start the playground:
 			</p>
 			<ServerState appState={appState} />
+			<p>
+				This Extension is based on https://github.com/WordPress/playground-tools/wp-now
+				so the default storage location is ~/.wp-now if there are any problems you can 
+				fix them via the wp-now cli or if you do not depend on the state even delete 
+				the directory it gets recreated.
```
danielbachhuber, line 66: Thanks for the PR, @lemanschik !

I think this is pretty specific information to include in this panel. Instead, what do you think about mentioning it in the project's README, and then adding a "Learn more" link here?

lemanschik: It needs to be in the panel for UX i added this info because i needed it most vscode users do not even know how to review the extension readme after install this information needs to be there when your using the extension as soon as the question comes up. 

eg installing none working plugins leads to never again working wp-now deployment manual deleting the plugin is needed and so on.

lemanschik: i would even prefer to add a whole open wp-now config folder button or virtual fs mount to speed up development.

lemanschik:
closes: 
- [ ] https://github.com/WordPress/playground-tools/issues/98

danielbachhuber:
@lemanschik This PR doesn't actually close that issue, FYI.

I don't think this documentation is quite the right approach for the issue. I'm going to close this PR in favor of https://github.com/WordPress/playground-tools/issues/104

Thanks for opening it, though!

PR #105, opened 2023-08-29
Require node v20 when `--blueprint=<file>` is used

### LINKED ISSUE #103: Incorrect required version
I tried to run wp-now with Node 16 and got this error message:

`This script is requires node version v18.0.0 or above; found v16.20.2`

I switched to Node 18 and tried to run the command with blueprint:

`wp-now start --blueprint=blueprint.json`

But got an error:

```
Starting the server......
directory: /srv/www/app
mode: plugin
php: 8.0
wp: latest
blueprint steps: 1
ReferenceError: CustomEvent is not definedReferenceError: CustomEvent is not defined
    at F.notifyProgress (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/node_modules/@wp-playground/blueprints/index.js:725:11)
    at F.finish (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/node_modules/@wp-playground/blueprints/index.js:664:165)
    at Object.run (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/node_modules/@wp-playground/blueprints/index.js:2017:11)
    at async yt (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/node_modules/@wp-playground/blueprints/index.js:2079:3)
    at async startWPNow (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/main.js:596:5)
    at async startServer (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/main.js:865:42)
    at async Object.handler (file:///home/myuser/.nvm/versions/node/v18.17.1/lib/node_modules/@wp-now/wp-now/main.js:1037:25)
Failed to start the server: CustomEvent is not defined
```

After some debugging I found out that CustomEvent is supported from Node 19 onwards:
https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent#browser_compatibility

I switched to Node 20 and the command worked.

Seems that the suggested Node version 18 is incorrect.

I'm using wp-now version 0.1.61.

danielbachhuber:
Fixes https://github.com/WordPress/playground-tools/issues/103

## What?

Updates `with-node-version.js` to require node v20 when the `--blueprint=<file>` flag is used.

## Why?

node v20 is required for Blueprints, but node v18 is required for wp-now. Without this logic, wp-now crashes on node v18 when the `--blueprint=<file>` argument is used.

## Testing Instructions

1. Switch to this branch.
2. Run `npx nx build wp-now` to build the compiled version.
3. Switch to node 14 and observe the following behavior:

```
$ node /path/to/dist/packages/wp-now/with-node-version.js start --php=8.1
This script is requires node version v18.0.0 or above; found v14.21.3
$ node /path/to/dist/packages/wp-now/with-node-version.js start --php=8.1 --blueprint=foo
This script is requires node version v20.0.0 or above; found v14.21.3
```

4. Switch to node v18.
5. Verify `wp-now start` works as expected, but `wp-now start --blueprint=foo` fails.
6. Switch to node v20.
7. Verify `wp-now start` works as expected with and without a `--blueprint=` argument.

cc @dmsnell

### DIFF
--- a/packages/wp-now/public/with-node-version.js
+++ b/packages/wp-now/public/with-node-version.js
@@ -6,11 +6,13 @@ import fs from 'fs';
 // Set the minimum required/supported version of node here.
 
 // Check if `--blueprint=` is passed in proccess.argv
-const hasBlueprint = process.argv.findIndex((arg) => arg.includes('--blueprint='));
+const hasBlueprint = process.argv.some((arg) =>
+	arg.startsWith('--blueprint=')
+);
 
 const minimum = {
 	// `--blueprint=` requires node v20
-	major: -1 !== hasBlueprint ? 20 : 18,
+	major: hasBlueprint ? 20 : 18,
 	minor: 0,
 	patch: 0,
 };
--- a/packages/wp-now/public/with-node-version.js
+++ b/packages/wp-now/public/with-node-version.js
@@ -4,8 +4,13 @@ import path from 'path';
 import fs from 'fs';
 
 // Set the minimum required/supported version of node here.
+
+// Check if `--blueprint=` is passed in proccess.argv
+const hasBlueprint = process.argv.findIndex((arg) => arg.includes('--blueprint='));
+
 const minimum = {
-	major: 18,
+	// `--blueprint=` requires node v20
+	major: -1 !== hasBlueprint ? 20 : 18,
 	minor: 0,
 	patch: 0,
 };
--- a/packages/wp-now/README.md
+++ b/packages/wp-now/README.md
@@ -61,7 +61,7 @@ wp-now php my-file.php
 -   `--php=<version>`: the version of PHP to use. This is optional and if not provided, it will use a default version which is `8.0`(example usage: `--php=7.4`);
 -   `--port=<port>`: the port number on which the server will listen. This is optional and if not provided, it will pick an open port number automatically. The default port number is set to `8881`(example of usage: `--port=3000`);
 -   `--wp=<version>`: the version of WordPress to use. This is optional and if not provided, it will use a default version. The default version is set to the [latest WordPress version](https://wordpress.org/download/releases/)(example usage: `--wp=5.8`)
--   `--blueprint=<path>`: the path of a JSON file with the Blueprint steps. This is optional, if provided will execute the defined Blueprints. See [Using Blueprints](#using-blueprints) for more details.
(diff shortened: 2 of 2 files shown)


packages/wp-now/public/with-node-version.js
```diff
@@ -4,8 +4,13 @@ import path from 'path';
 import fs from 'fs';
 
 // Set the minimum required/supported version of node here.
+
+// Check if `--blueprint=` is passed in proccess.argv
+const hasBlueprint = process.argv.findIndex((arg) => arg.includes('--blueprint='));
```
dmsnell: this is a boolean-named variable but stores an integer. also it's checking if the value of the arg includes the string, but we only want to match if it starts with the string, also I would think it could appear without the equals sign (something to verify). if we use `process.argv.some((arg) => arg.startsWith('--blueprint=')` it's a boolean and we won't have to leak the integer abstraction throughout the rest of the code below

dmsnell (commented):
thanks for the patch @danielbachhuber

I'm leaning towards pushing this through if we fix the mismatch between the name of the `hasBlueprint` binding and its value, but it also brings me back to that fundamental question: why support Node 18 at all, and if not, why not just polyfill.

I'm not thrilled about pushing out a package with a flowchart for its minimum supported runtime. then again, right now it just crashes, so I appreciate that this at least communicates better.

if we do this, we probably ought to communicate differently that Blueprints require Node 20 otherwise it's going to be confusing to be told on the CLI that two separate invocations require separate minimum Node versions.

packages/wp-now/public/with-node-version.js
danielbachhuber: I'll pass that feedback along to Copilot 🤭

Fixed with 838946b8

danielbachhuber:
> it also brings me back to that fundamental question: why support Node 18 at all, and if not, why not just polyfill.

@dmsnell I'm definitely open to that discussion. I don't have a strong opinion that we should support Node 18, but it seems like we should as long as it's not too much hassle.

I created an issue to discuss further: https://github.com/WordPress/playground-tools/issues/106

PR #110, opened 2023-09-14
Add pooling and request limiting to prevent memory leaks

seanmorris:
## What is this PR doing?

* Prevents PHP's memory leaks from causing the system to stop working once memory has been exhausted. It follows in the footsteps of systems like PHP-FPM by limiting the number of requests a PHP "thread" can handle before it is discarded and restarted, using Pooling behavior in `packages/wp-now/src/start-server.ts`.
* ~~`--inspect` and `--inspect-brk` switches have been added to allow the debugging of nodejs based WASM scripts in chrome dev tools.~~
* ~~`--trace-warnings`, `--trace-uncaught`, and `--trace-exit` flag have also been added to aid in debugging.~~ 

~~Struck~~ lines moved to https://github.com/WordPress/playground-tools/pull/132

## What problem is it solving?

PHP will eventually hit a memory leak when handling many requests one after another. 

## How is the problem addressed?

This PR adds behavior to track the number of requests a given PHP instance has handled, and discards it after a certain amount. A pool of PHP instances can be maintained so a fresh instance is always available, while discarded instances are replaced.

### in `wordpress-playground`

This requires the latest development version of `wordpress-playground`, so we'll need to check out the `trunk` branch and link some packages first:

```
git checkout trunk
nvm use
npm install
npm run build
```

Then, switch to the `dist/packages/playground/blueprints/` directory and link the package:

```
cd dist/packages/playground/blueprints/
npm link
```

In another terminal, switch to the `dist/packages/php-wasm/node/ ` directory and link that package as well:

```
cd dist/packages/php-wasm/node/ 
npm link
```

### In this project:

Run:

```bash
npm link @php-wasm/node @wp-playground/blueprints
npx nx build wp-now
node dist/packages/wp-now/cli.js start
```

Navigate to http://localhost:8881/

Set up auto refresh to refresh every few seconds and leave it running indefinitely.

You can use the this extenstion to do that in chrome: https://chromewebstore.google.com/detail/amazing-auto-refresh/lgjmjfjpldlhbaeinfjbgokoakpjglbn


packages/nx-extensions/src/executors/built-script/schema.json
```diff
@@ -10,6 +10,11 @@
 			"description": "Path of the script to run.",
 			"x-prompt": "What script would you like to run?"
 		},
+		"debug": {
+			"type": "boolean",
+			"description": "Use devtools as a debugger.",
+			"x-prompt": "Would you like to use devtools?"
```
dmsnell: this is a fairly nondescript prompt. it could help to guide people a bit more into what they may not know is there. something like "Would you like to connect a Node debugging client?"

"devtools" is probably a bit too casual as well, since there are plenty of non-devtools debuggers.

I wonder too if we would always want to use `--inspect-brk` or if there's value in running `--inspect` as well. if we used these, we could open both up and allow a closer correspondence to the Node terminology people may be used to.

packages/wp-now/src/config.ts
```diff
@@ -111,6 +114,7 @@ export default async function getWpNowConfig(
 		phpVersion: args.php as SupportedPHPVersion,
 		projectPath: args.path as string,
 		wordPressVersion: args.wp as string,
+		maxRequests: args.maxRequests as number,
```
dmsnell, line 120: what is the purpose in choosing the max request count? is there a compelling reason to configure this or would it be just as good for us to pick a reasonable value and leave it at that?

what are the tradeoffs involved?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,180 @@
...
+	notifiers = new Map(); // Inverted promises to notify async code of backlogged item processed.
+	running = new Set(); // Set of busy PHP instances.
+	backlog = []; // Set of request callbacks waiting to be run.
+
+	constructor({
+		spawner = async (): Promise<any> => {},
+		maxRequests = 2000,
+		maxJobs = 5,
+	} = {}) {
+		Object.assign(this, { spawner, maxRequests, maxJobs });
```
dmsnell: this is a style note, but I think we would generally prefer to set these like `this.spawner = spawner`

did you see this pattern elsewhere here?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,180 @@
...
+		this[Reap]();
+		this[Spawn]();
```
dmsnell: what's the purpose of using these symbols instead of simply using `reap`, `spawn`, and `fatal` as names?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,180 @@
...
+				this[Spawn]();
+
+				if (!this.backlog.length) {
+					return;
+				}
+
+				const idleInstanceNext = this.getIdleInstance();
+
+				const next = this.backlog.shift();
+				const info = this.instances.get(idleInstanceNext);
```
dmsnell: couldn't `ideInstanceNext` here be false too?

dmsnell (commented):
this is looking good.

 - there's use of the `/**` JSDoc comment opener, but we aren't particularly following JSDoc conventions. it would be nice if we were consistent here.
 - do you think this same issue is likely present in the web version of the Playground? maybe it's rarer since people won't keep it open as long as they keep `wp-now` running? is there an easy way you can think of to refresh the PHP instance from within the Playground itself so that tools like `wp-now` don't also need to apply this kind of fix?

packages/wp-now/src/pool.ts
seanmorris: While this function is kicked off by the resolution of a previous promise that SHOULD free up the instance its using, I could see how it would be possible for a FALSE to come in here if we get a request at JUST the wrong millisecond.

This is a little complicated because we're actually managing TWO pools here, one of backlogged requests and one of idle/running PHP instances, but I think I can come up with something robust to that case.

packages/wp-now/src/pool.ts
seanmorris: That's just a pattern I use for private methods.

packages/wp-now/src/pool.ts
seanmorris: That's just something I use as shorthand for multiple assignments.

packages/nx-extensions/src/executors/built-script/schema.json
seanmorris: You're definitely right. Keeping the params consistent would be a good idea. I also plan to add `--trace-warnings`.

packages/wp-now/src/config.ts
seanmorris: Depending on how php is compiled/utilized, its possible for users to have issues at different levels of process-reuse. To account for that, I'd like to allow our users to tune this parameter to their own liking.

seanmorris: One of the main tradeoffs is that during the switch from one instance to the next there seems to be some issue with keeping the cookies synced up correctly. The issue quickly resolves itself but will result in a couple of 403's when the handoff happens. The solution would be to find out why the instances aren't sharing sessions and correct that.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,261 @@
...
+				info.requests++;
+
+				let request;
+
+				try {
+					// Don't ACTUALLY do anything until the
+					// instance is done spawning.
+					request = next(await nextInstance);
+				}
+				catch(error) {
```
seanmorris: @dmsnell Let me know what you think of this logic. The alternative to re-queuing after a failed spawn is to return a 502 bad gateway.

Interested to know your thoughts.

packages/wp-now/src/config.ts
dmsnell: > The issue quickly resolves itself but will result in a couple of 403's when the handoff happens

This seems like a blocking problem, a kind of major issue.

packages/wp-now/src/pool.ts
dmsnell: I'm a bit confused on what happens in these cases and probably need to think about it more. One thing that worries me is that if there is something intrinsic in the request that led to the failure, and we `unshift` it to the beginning of the queue, could we get into situations that lead to infinite and infinitely-fast re-death of newly spawned processes. I'm not sure if that worry is sound though, as I don't fully understand what's happening with these workers. It's not like they are OS processor or have the same characteristics of overhead and startup time.

In general, a `500` seems like an acceptable response. Why specifically a 502? The lines here are blurry.

packages/wp-now/src/wp-now.ts
```diff
@@ -86,13 +109,31 @@ export default async function startWPNow(
...
-			runIndexMode(_php, options);
-		});
-		return { php, phpInstances, options };
+		const spawnAndSetup = async () => {
+			const instance = await spawnInstance();
+			await runIndexMode(instance, options);
+			return instance;
+		};
+
+		await phpInstances.map((instance) => runIndexMode(instance, options));
```
seanmorris: Note to self: Wrap with Promise.all()

packages/wp-now/src/config.ts
seanmorris: @dmsnell Fixed, requires https://github.com/WordPress/wordpress-playground/pull/639

seanmorris:
> this is looking good.
> 
> * there's use of the `/**` JSDoc comment opener, but we aren't particularly following JSDoc conventions. it would be nice if we were consistent here.
> * do you think this same issue is likely present in the web version of the Playground? maybe it's rarer since people won't keep it open as long as they keep `wp-now` running? is there an easy way you can think of to refresh the PHP instance from within the Playground itself so that tools like `wp-now` don't also need to apply this kind of fix?

The comment style is a force of habit, I can look over the existing style and make it consistent.

If this code need to be upstreamed into wordpress-playground, that is 100% possible. The Pool class is not coupled to anything in particular, so we can absolutely provide it out of the box in wordpress-playground.

packages/wp-now/src/pool.ts
seanmorris: If I recall correctly an HTTP daemon will serve a 502 if it can't properly instantiate PHP to handle a request, but a 500 seems valid as well.

adamziel:
Is this ready for another review now?

seanmorris:
> Is this ready for another review now?

@adamziel  Sure! I'm excited to hear your thoughts.

packages/nx-extensions/src/executors/built-script/executor.ts
```diff
@@ -7,6 +7,11 @@ const dirname = __dirname;
 
 export default async function runExecutor(options: BuiltScriptExecutorSchema) {
 	const args = [
+		...(options['inspect'] ? ['--inspect-brk'] : []),
```
adamziel: For posterity – let's find a way to reuse `nx-extensions` between `playground-tools` and `playground` repos.

adamziel: I opened https://github.com/WordPress/wordpress-playground/issues/650 to track this

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,261 @@
...
+
+/**
+ * PRIVATE
+ * Spawns new instances if the pool is not full.
+ * Returns a list of new instances.
+ */
+const spawn = (pool) => {
+	const newInstances = new Set();
+
+	while (pool.maxJobs > 0 && pool.instanceInfo.size < pool.maxJobs) {
```
adamziel: Nitpick: pool.maxJobs doesn't change in the loop and makes the condition a bit more difficult to follow. How about`if ( pool.maxJobs <= 0 ) { return newInstances; }` before the loop?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,261 @@
...
+		const info = new PoolInfo();
+		const instance = pool.spawner();
```
adamziel: What do you think about `spawn()` as the name here instead of `spawner()`?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,261 @@
...
+		return instance;
+	}
+
+	return false;
+};
+
+/**
+ * Maintains and refreshes a list of php instances
+ * such that each one will only be fed X number of requests
+ * before being discarded and replaced.
```
adamziel, line 111: Do you have any insights about the root cause of the leak? I don't think `php-fpm` needs to do the same kind of pooling&discarding so it must be something on our side. Emscripten issue perhaps?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,261 @@
...
+ */
+export class Pool {
+	instanceInfo = new Map(); // php => PoolInfo
+
+	spawner: () => Promise<any>; // Callback to create new instances.
+	maxRequests: number; // Max requests to feed each instance
+	maxJobs: number; // Max number of instances to maintain at once.
+
+	notifiers = new Map(); // Inverted promises to notify async code of backlogged item processed.
```
adamziel: Nit: I had to check with `enqueue` what is this doing as I didn't understand the name and the comment. Perhaps `resolvers` would be more descriptibe, since `new Promise(resolve => { ... });` is a popular pattern 🤔

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,261 @@
...
+			// rejected later when the item is processed.
+			const notifier = new Promise((accept, reject) =>
+				this.notifiers.set(item, { accept, reject })
+			);
+
+			// Return the notifier so async calling code
+			// can still respond correctly when the item
+			// is finally processed.
+			return notifier;
+		} else {
```
adamziel: Nit: No need for the `else`

adamziel:
How much blocking time does spawning a new PHP instance take? From `wp-now` perspective, would it make sense to spawn a whole new node process?

packages/wp-now/src/pool.ts
seanmorris: @adamziel 
Originally, this function (`const spawn = (pool) => {...}`) WAS `Pool.spawn`, until I made it private and remove it from the object, so I had to use a different name for the provided callback.

Now that I've refactored it into a private function, its completely possible for me to use the correct, verb based name for the function.

packages/wp-now/src/pool.ts
seanmorris: @adamziel That would definitely be clearer. I'll go ahead and make the change.

packages/wp-now/src/pool.ts
seanmorris: https://www.php.net/manual/en/install.fpm.configuration.php#pm.max-requests

This pathological behavior does exist in native PHP, but won't cause problems nearly as quickly since its not restricted to a single gigabyte of memory. 

Emscripten definitely introduces unique challenges. Since we're dealing with a linear, "physical" memory array, as opposed to a virtual memory system afforded by most modern OSes, we're prone to things like memory fragmentation. In that situation, we could have the entire gigabyte empty except for a few sparse allocations. If no **_contiguous_** region of memory exists for the length requested, memory allocations will fail. This tends to happen when a new request attempts to initialize a heap structure but cannot find a contiguous  2mb chunk of memory.

We can go as far as debugging PHP itself, and contributing the fix upstream. But even in this case we cannot guarantee that a third party extension will not introduce a leak sometime in the future.

Therefore, we should have a solution robust to memory leaks that come from upstream code. I think that following the native strategy is the best way.

packages/nx-extensions/src/executors/built-script/executor.ts
seanmorris: @adamziel What do you think of this: We can make it its own NPM package, and the use a relative symlink to take the directory in `node_modules/` and put it wherever we need in the folder structure.

Git & GitHub both handle relative symlinks pretty well, and if I remember correctly windows-based git clients will translate it to/from native "shortcuts" transparently.

packages/wp-now/src/pool.ts
seanmorris: These nitpicks tell me you're looking at my code closely.

I appreciate it, and I'll go ahead and make that change.

packages/wp-now/src/pool.ts
seanmorris: @adamziel Yea, this is definitely a weird pattern. I've only seen it implemented in a handful of places in my career. I like the name `resolvers` for this field. Its definitely more descriptive.

adamziel: Thanks!

I saw a name `PromiseQueue` used for this general pattern. There was even a generic implementation used in Playground some months back but it was removed and simplified later on.

packages/wp-now/src/pool.ts
adamziel: Oh of course, third party extensions! Sean, that's such a good answer, thank you. I'd love to see your GitHub comment as a part of this PR in an actual comment block – it clearly explains the rationale.

> We can go as far as debugging PHP itself, and contributing the fix upstream. But even in this case we cannot guarantee that a third party extension will not introduce a leak sometime in the future.

Agreed, let's just stick with discarding PHP instances here.

packages/nx-extensions/src/executors/built-script/executor.ts
adamziel: I'd like to avoid a workflow where testing changes between repos requires pushing something to GitHub – would that do it? If not, then I wonder if there's a solution where we could just make a change in `nx-extensions` in a local copy of the `wordpress-playground` repo and have it automatically reflected in the local copy of the `playground-tools` repo. CC @dmsnell who spent a ton of time exploring a similar problem in context of Gutenberg and Tumblr.

packages/wp-now/src/config.ts
adamziel: I'm not a big fan of exposing that setting, but I also don't see any way to automatically use a single optimal value as that will depend on many variables like hardware. Yup, let's just expose it.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * PRIVATE
+ * Reaps children if they've passed the maxRequest count.
+ */
+const reap = (pool) => {
+	for (const [instance, info] of pool.instanceInfo) {
+		if (pool.maxRequests > 0 && info.requests >= pool.maxRequests) {
+			info.active = false;
+			pool.instanceInfo.delete(instance);
+			instance.then((unwrapped) => unwrapped.exit()).catch((error) => {});
```
adamziel:
```suggestion
			instance.then((unwrapped) => unwrapped.exit()).catch(() => {});
```

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,263 @@
...
+	async enqueue(item: (php: NodePHP) => Promise<any>) {
+		const idleInstance = getIdleInstance(this);
+
+		if (!idleInstance) {
+			// Defer the callback if we don't have an idle instance available.
+			this.backlog.push(item);
+
+			// Split a promise open so it can be accepted or
+			// rejected later when the item is processed.
+			const notifier = new Promise((accept, reject) =>
```
adamziel: Since it's `resolvers` now, it seems natural to rename accept to resolve.

adamziel: Oh, I see, it's also used in a few other places around this class. It would be lovely to use the same words to refer to the same concepts, e.g. `notifier` here is a Promise, but in line 205 a `notifier` is an `{accept, reject}` object. Since it comes from `resolvers`, the name in like 205 could be `resolver`.

I know these wording changes may seem minor, but naming consistency goes a long way for a new person or when debugging a tricky issue. For example, [this Gutenberg PR](https://github.com/WordPress/gutenberg/pull/39349) saved me a ton of hours down the road.

packages/nx-extensions/src/executors/built-script/executor.ts
seanmorris: @adamziel No need to commit, we'd use `npm link` for that.

https://docs.npmjs.com/cli/v9/commands/npm-link

This allows you to take any package and symlink it INTO the `node_modules/` directory of another project on the same system, so all the updates are available as you make your changes.

The symlink that lives at `packages/nx-extensions` won't care if it points at a directory or another symlink.

adamziel: That makes a lot of sense, thank you @seanmorris! Let's proceed with this PR as it is and then use #650 to track code reuse further down the road

packages/wp-now/src/pool.ts
seanmorris: @adamziel  https://github.com/WordPress/playground-tools/pull/110/files#diff-95251f05603f485be0377cd57ce90fe740821ca840c6053a115cb48c8725341bR103

packages/wp-now/src/pool.ts
seanmorris: @dmsnell @adamziel Lets revisit this, I don't want to forget about it.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,285 @@
...
+	 * @param options - {spawn, maxRequests, maxJobs}
+	 */
+	constructor({
+		maxRequests = 128,
+		maxJobs = 1,
+		spawn = async (): Promise<any> => {},
+		fatal = (instance: instance, request: request) => {},
+		reap = (instance: instance) => {},
+	} = {}) {
+		Object.defineProperties(this, {
```
seanmorris, line 158: @dmsnell These props are now read-only. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties

seanmorris:
@sejas Would you mind reviewing my PR and letting me know if it gets your stamp of approval?

packages/wp-now/src/wp-now.ts
```diff
@@ -141,23 +186,33 @@ export default async function startWPNow(
 	}
 
 	await installationStep2(php);
+
 	await login(php, {
 		username: 'admin',
 		password: 'password',
 	});
 
+	const initialCookies = php.requestHandler.serializeCookies().split(';');
```
sejas, line 168: I'm getting an error that stops me to run `wp-now start`.
@seanmorris , is there something I'm doing wrong?

Here is my error:

```
node dist/packages/wp-now/cli.js start                                                                                   Node 20.8.1 13:44:19
Starting the server......
directory: /Users/macbookpro/proyectos/a8c/playground-tools
mode: playground
php: 8.0
wp: latest
TypeError: php.requestHandler.serializeCookies is not a function
    at startWPNow (file:///Users/macbookpro/proyectos/a8c/playground-tools/dist/packages/wp-now/main.js:851:45)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async startServer (file:///Users/macbookpro/proyectos/a8c/playground-tools/dist/packages/wp-now/main.js:1118:48)
    at async Object.handler (file:///Users/macbookpro/proyectos/a8c/playground-tools/dist/packages/wp-now/main.js:1325:25)
Failed to start the server: php.requestHandler.serializeCookies is not a function
```

**Test instructions**
1. Run `nvm use v20`
2. Run `npx nx build wp-now`
3. Run `node dist/packages/wp-now/cli.js start`
4. Observe the mentioned error

sejas (commented):
@seanmorris, the general approach seems correct to me. I think the idea of the Pool is great. In my initial implementation I had something similar to run some commands before exposing the port to the user. 

https://github.com/WordPress/playground-tools/blob/ef3c6d030bac39502664f1514acba3c21576d9b1/packages/wp-now/src/wp-now.ts#L69-L74

I didn't approve the PR yet, because I couldn't run the project, is there something I'm missing? I added the error output in a comment below.

packages/wp-now/src/wp-now.ts
seanmorris: @sejas We just deployed a new version of `wordpress-playground` that exposes the `serializeCookies` and `exit` methods, which were previously private. Try updating that package and see if the error goes away. Sorry if I'm less than responsive today, its a holiday in the states and I've got to run to see family. 

https://docs.npmjs.com/cli/v6/commands/npm-link

adamziel: The new packages aren't released yet but I just triggered the release workflow so it should be soon.

It sounds like this PR should ship a package.json update to use the latest packages? Although that might depend on https://github.com/WordPress/playground-tools/pull/116 which may need to land first.

seanmorris: @adamziel Thanks for chiming in! I had assumed that merging into `trunk` would publish an update automatically. In the meantime, is it possible to pull directly from `trunk` via npm? If not, I know its possible to use `npm link` to bring the latest version in.

adamziel: Publishing to trunk doesn't release new packages ATM. There's a GitHub action we're running manually from time to time.

packages/nx-extensions/src/executors/built-script/schema.json
seanmorris: @dmsnell ~~How does this look? https://github.com/WordPress/playground-tools/pull/110/files#diff-7216652cb6324a396ee2104b76f2e83e8b5e0aa6650acf7e6823d49b7a095078R16~~

This has been moved to:
https://github.com/WordPress/playground-tools/pull/132

packages/wp-now/src/pool.ts
seanmorris: @dmsnell This has been refactored: https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L160-L166

packages/wp-now/src/pool.ts
seanmorris: @dmsnell This has been refactored: https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L36-L85

packages/wp-now/src/pool.ts
seanmorris: @adamziel @dmsnell Refactored to return the error to calling code: https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L267-L273

packages/wp-now/src/pool.ts
seanmorris: If I remember correctly, removing that makes TypeScript complain. I can double check if need be.

packages/wp-now/src/pool.ts
seanmorris: @adamziel https://github.com/WordPress/playground-tools/pull/110/files#diff-95251f05603f485be0377cd57ce90fe740821ca840c6053a115cb48c8725341bR137

adamziel:
What would be the next step here?

seanmorris:
> What would be the next step here?

I think all that's left to do is get everything ready for release and start incrementing version numbers.

packages/interactive-code-block/project.json
```diff
@@ -14,7 +14,7 @@
 				"cwd": "dist/packages",
 				"commands": [
 					"cp ../../packages/interactive-code-block/README.md ./interactive-code-block/",
-					"rm interactive-code-block/assets/php_5*",
+					"rm -f interactive-code-block/assets/php_5*",
```
adamziel: Why only add `-f` in this line of code but not the other ones? And why update the `interactive-code-block`?

adamziel:
> --inspect and --inspect-brk switches have been added to allow the debugging of nodejs based WASM scripts in chrome dev tools.

How can those switches be used to debug wp-now?

packages/vscode-extension/project.json
```diff
@@ -46,7 +46,7 @@
 					"cp ./LICENSE dist/packages/vscode-extension",
 					"cp packages/vscode-extension/package.json dist/packages/vscode-extension",
 					"cp packages/vscode-extension/README.md dist/packages/vscode-extension",
-					"cp node_modules/@php-wasm/node/*.wasm dist/packages/vscode-extension"
+					"find node_modules/@php-wasm/node/ -type f -name php_*.wasm | while read FILE; do cp $FILE dist/packages/vscode-extension; done"
```
adamziel: Why update the vscode-extension code in this PR? It seems like an unrelated change as this PR is about `wp-now`.

packages/wp-now/src/node-pool.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type nodePoolOptions = {
+	spawn?: () => Promise<NodePHP>;
+	reap?: (instance: NodePHP) => void;
+	fatal?: (instance: NodePHP, error: any) => any;
+	maxRequests?: number;
+	maxJobs?: number;
+};
+
+const defaultSpawn = async () => await NodePHP.load('8.2');
```
adamziel: Let's avoid hardcoding specific PHP versions and use the latest supported version. `LatestSupportedPHPVersion` should be available as an export in one of the `@php-wasm` packages.

packages/wp-now/src/node-pool.ts
```diff
@@ -0,0 +1,44 @@
...
+				error && error.stack ? error.stack : error
+			)}`
+		)
+	);
+
+const defaultReap = (instance: NodePHP) => {
+	try {
+		instance.exit();
+	} catch {
+		void 0;
```
adamziel: Why `void 0;`? It's quite unexpected and makes the reader stop to think "wait, what's going on here?". If that's just to make the linter happy, you could say `// Do nothing`.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,295 @@
...
+};
+
+let childCount = 0;
+
+/**
+ * Tracks stats of instances in a pool.
+ * @private
+ */
+class PoolInfo {
+	id = childCount++; // Unique ID for debugging purposes.
```
adamziel: This syntax makes it quite confusing whether the `childCount++` happens once and is a static `0`, or is it computed for every new instance. I had to check in my devtools because I didn't know, and I saw it's the latter. Let's just define an explicit constructor to set any computed values, and save the future readers confusion.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	return pool.fatal(instance, error);
+};
+
+/**
+ * Find the next available idle instance.
+ * @private
+ */
+const getIdleInstance = (pool) => {
```
adamziel: Let's set the type of `pool`

packages/wp-now/src/wp-now.ts
```diff
@@ -67,11 +75,26 @@ export default async function startWPNow(
...
+
+	const spawnInstance = async () => {
+		const php = await NodePHP.load(options.phpVersion, nodePHPOptions);
+
+		php.mkdirTree(documentRoot);
+		php.chdir(documentRoot);
+		php.writeFile(
+			`${documentRoot}/index.php`,
+			`<?php echo 'Hello wp-now!';`
+		);
```
adamziel: It looks like a dev artifact, but at a risk of stating the obvious – let's not overwrite the index.php file when spawning a PHP instance.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,295 @@
+type instance = any;
+
+type request = (instance: instance) => Promise<any>;
+
+export type poolOptions = {
```
adamziel: Let's capitalize all type names – that's the convention used everywhere in Playground, and I think also mostly in TypeScript (except for primitives).

```suggestion
export type PoolOptions = {
```

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,295 @@
+type instance = any;
```
adamziel: I immediately have a question in my head: "an `instance` of what?". Let's include the answer in the type name.

packages/wp-now/src/pool.ts
adamziel: Which comment is this a reply to? I'd love to learn more about the context here.

Also, Are you sure these are read-only now? I don't see any mention of `defineProperties` marking props as read-only by default on that documentation page. Also, that would be a useful comment to include in the code with a rationale why it's important for them to be read-only.

adamziel: Aha, if the `value` is defined, but `writable` isn't, then `writable` defaults to false. That's not obvious from the code unless you know the JavaScript doc by heart, and even if you do it's still easy to miss. Let's make that code very obvious and save the reader some confusion and a round-trip to MDN, and explicitly state `writable: false`.

packages/interactive-code-block/project.json
seanmorris: I was getting an error during the build process due to the attempted removal of a non-existent file. `-f` prevents `rm` from throwing errors when you try to remove a file that isn't there. I've reverted it locally and the issue doesn't seem to be happening anymore. I only applied this to php5 because I've seen that we're deprecating that.

packages/vscode-extension/project.json
seanmorris: @adamziel This was originally a companion PR to https://github.com/WordPress/wordpress-playground/pull/639, and when that project is built with sourcemaps, the WASM files appear in subpaths, along with sourcemaps and sourcetrees. This ensures the subpaths are copied over.

If that's not relevant for the context of the extension, I can remove it.

packages/wp-now/src/node-pool.ts
seanmorris: @adamziel I was thinking about this, but if wp-now is started with a different PHP version, I think we should continue to use that. Do you know how I can grab the value passed on the command line here?

packages/wp-now/src/node-pool.ts
seanmorris: I didn't realize a comment would be enough to quiet it down, I'll correct that.

packages/interactive-code-block/project.json
adamziel: Aha! PHP is not in the Playground repo anymore so that entire `rm` line could be removed in a separate PR.

packages/vscode-extension/project.json
adamziel: Yeah it sounds relevant to #639, just not to pooling and request limiting this PR is dealing with. Would you please file an issue to keep track of this problem separately?

packages/wp-now/src/node-pool.ts
seanmorris: @adamziel Ah wait, this is just the default spawner, the calling code handles that.

adamziel: > but if wp-now is started with a different PHP version, I think we should continue to use that
> Do you know how I can grab the value passed on the command line here?

I'm not sure I follow, it seems like `wp-now` provides a custom `spawn` callback that uses the PHP version passed as the CLI arg?

packages/wp-now/src/pool.ts
seanmorris: @adamziel  https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L19-L28

packages/wp-now/src/pool.ts
seanmorris: @adamziel https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L91

packages/wp-now/src/wp-now.ts
seanmorris: @adamziel https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/wp-now.ts#L85-L87

packages/wp-now/src/pool.ts
seanmorris: @adamziel https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L5

packages/wp-now/src/pool.ts
seanmorris: @adamziel https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/pool.ts#L1

seanmorris:
> > --inspect and --inspect-brk switches have been added to allow the debugging of nodejs based WASM scripts in chrome dev tools.
> 
> How can those switches be used to debug wp-now?

They were *originally* for debugging php-wasm itself, but can be used to debug javascript as well. You can set breakpoints and inspect the memory in dev tools. When you've built the wordpress-playground package with sourcemaps, and linked it, they allow you to set breakpoints in the C code, using devtools.

Before you comment, I'm going to guess you're gonna ask me to move this into its own PR?

https://github.com/WordPress/playground-tools/pull/132

packages/interactive-code-block/project.json
seanmorris: @adamziel See, that worries me. If someone still has that directory, and then moves to a version that doesn't have an `rm` call for it, then it won't be removed. I think we should **definitely** open the PR now, **BUT** I think we should hold off on actually merging it for a predetermined amount of time, so that this can run off. Let me know your thoughts.

packages/vscode-extension/project.json
seanmorris: @adam https://github.com/orgs/Automattic/projects/821/views/3?pane=issue&itemId=46957567

packages/interactive-code-block/project.json
seanmorris: https://github.com/WordPress/playground-tools/pull/131

packages/wp-now/src/node-pool.ts
seanmorris: @adamziel Yep, the calling code provides a spawner. I just didn't have everything in front of me when I wrote that first comment.

packages/interactive-code-block/project.json
adamziel: How would someone still have that directory after `git pull` and `npm install`?

adamziel:
I saw this came up in the review queue, but there are some failing CI checks and open conversations so it seems like it's not ready for a review yet.

seanmorris:
> I saw this came up in the review queue, but there are some failing CI checks and open conversations so it seems like it's not ready for a review yet.

@adamziel That's going to fail until we plan a new release of `wordpress-playground`. The relevant code is already merged into `trunk`:

https://github.com/WordPress/wordpress-playground/blob/7ec4b9a0bc014baf69c3f53f1f59b40816a3aed0/packages/php-wasm/universal/src/lib/base-php.ts#L681-L683

https://github.com/WordPress/wordpress-playground/blob/7ec4b9a0bc014baf69c3f53f1f59b40816a3aed0/packages/php-wasm/universal/src/lib/php-browser.ts#L115-L138

packages/interactive-code-block/project.json
seanmorris: @adamziel If someone pulled the package freshly then that directory wouldn't exist. But if someone were already working with it, and had it in a state where that directory exists, and then checked out a version where the `rm` call is gone, then the directory might persist unexpectedly.

packages/wp-now/src/node-pool.ts
seanmorris: https://github.com/WordPress/playground-tools/blob/75bd3eb5d3d07a3dc17cd2c8373cd32000c68035/packages/wp-now/src/node-pool.ts#L13

packages/wp-now/src/wp-now.ts
```diff
@@ -44,34 +46,32 @@ async function applyToInstances(phpInstances: NodePHP[], callback: Function) {
...
+	options: WPNowOptions;
+	pool: NodePool;
+}> {
 	const { documentRoot } = options;
+
 	const nodePHPOptions: PHPLoaderOptions = {
 		requestHandler: {
 			documentRoot,
 			absoluteUrl: options.absoluteUrl,
-			isStaticFilePath: (path) => {
```
adamziel, line 53: Removing this triggers a lint error above – `seemsLikeAPHPFile` is now dead code.

packages/wp-now/src/tests/wp-now.spec.ts
```diff
@@ -859,4 +859,4 @@ describe('wp-cli command', () => {
 		await executeWPCli(['cli', 'version']);
 		expect(output).toMatch(/WP-CLI (\d\.?)+/i);
 	});
-});
+}, 30_000);
```
adamziel, line 862: Why increase this value?

packages/wp-now/src/wp-now.ts
seanmorris: Strange, I was getting linter errors on my end as if it were deprecated. That's what prompted me to remove it to begin with.

I'll double check that.

packages/vscode-extension/project.json
seanmorris: @adamziel The change has been reverted in this branch.

packages/wp-now/src/wp-now.ts
seanmorris: @adamziel Looks like that type doesn't allow for the `isStaticFilePath` key. Is the `isStaticFilePath` key not expected to be removed? If I'm missing something, lets revisit this one tomorrow morning.

GitHub search for `isStaticFilePath`: https://github.com/search?q=org%3AWordPress+isStaticFilePath&type=code

Current Commit (`30a561d97bdd7dee6674528dfff95c912bc734d3`) https://github.com/WordPress/wordpress-playground/blob/30a561d97bdd7dee6674528dfff95c912bc734d3/packages/php-wasm/universal/src/lib/php-request-handler.ts#L17-L27

Trunk: https://github.com/WordPress/wordpress-playground/blob/trunk/packages/php-wasm/universal/src/lib/php-request-handler.ts#L17-L27

packages/wp-now/src/pool.ts
seanmorris: https://github.com/WordPress/playground-tools/blob/f352bb84c1c67a45bbb27434c19b463f7a33b908/packages/wp-now/src/pool.ts#L160-L191

packages/wp-now/src/pool.ts
seanmorris: https://github.com/WordPress/playground-tools/blob/f352bb84c1c67a45bbb27434c19b463f7a33b908/packages/wp-now/src/pool.ts#L110-L131

packages/wp-now/src/tests/wp-now.spec.ts
seanmorris: I was getting test-timeouts with the default.

packages/wp-now/src/wp-now.ts
seanmorris: @adamziel I noticed you published a new version! I wish you'd told me! I'd have gotten these linter errors out of the way immediately. Thank you for getting that going though, I jumped on this as soon as I noticed.

packages/wp-now/src/wp-now.ts
adamziel: It is expected to be removed, you'll just need to remove this now-unused function, too:

CleanShot 2023-12-22 at 13 18 27@2x

packages/wp-now/src/wp-now.ts
adamziel: >  I noticed you published a new version! I wish you'd told me! 

I published the packages two days ago and let know on slack yesterday.

packages/wp-now/src/pool.ts
adamziel: @seanmorris I'd still like to learn more about the context here (as I asked in the comment above). What is this a reply to, and why do these properties need to be read only?

packages/interactive-code-block/project.json
adamziel: The way I understand git is that checking out a newer revision where a file is removed does remove that file from the filesystem. Am I getting that wrong?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,324 @@
...
+
+let childCount = 0;
+
+/**
+ * Tracks stats of instances in a pool.
+ * @private
+ */
+class PoolInfo {
+	id: number; // Unique ID for debugging purposes.
+	started: number; // Time spawned.
```
adamziel: This seems to be unused – let's remove it.

packages/wp-now/src/pool.ts
adamziel: Thanks! By the way, why `any`? `@php-wasm/node` and `@php-wasm/universal` ship types like `UniversalPHP` or `NodePHP` – would either be applicable here?

dmsnell:
Looking through this I didn't see any kind of manual memory allocation for spawning a new process, but I didn't check _behind the diff_ to see if that's there or if the question even makes sense.

Still, I was thinking about the comment about the linear memory and fragmentation (thanks for adding that) and wondered if we have any control over this, and whether we could add in a later PR a custom allocator for each PHP process so we could ensure reuse and a memory limit. E.g. if we set PHP's max memory, we could preallocate a multiple of that for every worker we build, then let new workers use the now-reaped memory from the previous worker.

packages/wp-now/src/pool.ts
seanmorris: @adamziel It doesn't really make sense to change these values, given the current pooling implementation. This all originally started based on the notation I was using to create private properties. Do you see a reason that calling code should be able to modify these values?

packages/wp-now/src/wp-now.ts
seanmorris: @adamziel Not sure how I missed that one, but I'm glad we got it published!

packages/wp-now/src/wp-now.ts
seanmorris: @adamziel That's been removed.

packages/wp-now/src/config.ts
```diff
@@ -43,6 +45,8 @@ export interface WPNowOptions {
 	wpContentPath?: string;
 	wordPressVersion?: string;
 	numberOfPhpInstances?: number;
+	maxRequests?: number;
+	maxJobs?: number;
```
dmsnell, line 49: what's the interplay here between `maxJobs` and `numberOfPhpInstances`?
we didn't recreate the same config value did we?

could we use `numberOfPhpInstances` to set `maxJobs` or are these somehow indicating different things: number of workers in a pool vs. number of pools? would it make sense to only use one and then to leave the existing config for `WPNowOptions`?

packages/wp-now/src/config.ts
```diff
@@ -20,6 +20,8 @@ export interface CliOptions {
 	port?: number;
 	blueprint?: string;
 	reset?: boolean;
+	maxRequests?: number;
+	maxJobs?: number;
```
dmsnell, line 24: I know it's late in the process but now as this is basically ready I'm questioning how clear these names are. we're not limiting the total number of requests, and what is a job?

what if we either renamed them or created them as sub-options like `poolOptions`?

something like `maxRequestsPerWorkerLifespan` is verbose but does far more to communicate the role of the config, and `maxPhpWorkers` (or even the already existing `numberOfPhpInstances`) skips the term for one that's already in scope when creating one of these config values.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+let childCount = 0;
+
+/**
+ * Tracks stats of instances in a pool.
+ * @private
+ */
+class PoolInfo {
+	id: number; // Unique ID for debugging purposes.
+	requests = 0; // Total requests processed.
+	active = false; // Whether instance is considered active.
```
dmsnell, line 22: it seems clear that `active` conveys some sense of being `active`, but I am confused on two things:

 - what does "active" mean? e.g. "// An instance is considered active if it's currently handling a request."
 - this is a property of a `PoolInfo` class, so which _instance_ is considered active? we've been discussing `PhpInstance` and other instances. should this be "Whether this pool is [whatever active means]"?

as noted in the next comment, we would get a lot for almost no effort by moving these into JSDoc comments

dmsnell, line 22: since we're already explaining these, let's use JSDoc syntax so that the comments follow the values.

```ts
/** Unique ID for debugging purposes. */
id: number;

/** Total requests processed by this pool/instance/something? */
requests = 0;

/** This pool is considered active if X Y Z */
active = false;
```

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+		pool.instanceInfo.set(instance, info);
+		info.active = true;
+		newInstances.add(instance);
+	}
+
+	return newInstances;
+};
+
+/**
+ * Reaps children if they've passed the maxRequest count.
```
dmsnell, line 51: this is a description of what the code currently does, but it would be less fragile as a comment if we described why it does what it does.

something like this…

```ts
/**
 * Reaps (terminates) all of the PHP instances
 * in a given pool that are deemed "too old."
 *
 * In order to prevent memory leaks inherent in
 * a long-living PHP worker, this function reaps
 * PHP instances after they have handled more
 * than the configured number of maximum requests.
 *
 * There could be different conditions for reaping,
 * but the intent is to limit the extent of a
 * memory leak and ensure any proper cleanup happens
 * even if a process goes astray.
 *
 * @private
 *
 * @param pool Reap old PHP instances in this pool.
 */
const reap = (pool: Pool) => { … }
```

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+		}
+
+		if (!info.active) {
+			continue;
+		}
+		return instance;
+	}
+
+	return false;
+};
```
dmsnell, line 106: I don't want to nitpick this because the function seems fine, but if you happen to have some spare cycles and want to rethink this function, it seems like the sort isn't necessary. if we're going to iterate through the pool to find the instance with the fewest requests, we can ensure we only ever iterate once by scanning with an accumulator, instead of potentially iterating multiple times through the sort and subsequence iteration.

```ts
const getIdleInstance = (pool: Pool) => {
	let minRequests  = Infinity;
	let idleInstance = null;

	for (const [instance, info] of pool.instanceInfo) {
		…

		if (info.requests < minRequests) {
			minRequests  = info.requests;
			idleInstance = instance;
		}
	}

	return idleInstance ?? false;
}
```

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+
+	spawn: () => Promise<any>; // Async callback to create new instances.
+	fatal: (instance: PhpInstance, error: any) => any; // Callback called on instance fatal errors.
+	reap: (instance: PhpInstance) => void; // Callback called on destroyed instances.
+	maxRequests: number; // Max requests to feed each instance
+	maxJobs: number; // Max number of instances to maintain at once.
+
+	resolvers = new Map(); // Inverted promises to notify async code of backlogged item processed.
+	running = new Set(); // Set of busy PHP instances.
+	backlog = []; // Set of request callbacks waiting to be run.
```
dmsnell, line 141: same note here on the comments. these are nice to have, but we're so close to JSDoc and if we swap then the comments will follow the variables around the code in an IDE.

packages/wp-now/src/pool.ts
dmsnell: these seem eager, but fine. if we find a reason to change them we can adjust them here. I'd be happy either way, as we could also make them normal properties and lock them down should we discover the need. 🤷‍♂️

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+				configurable: false,
+				enumerable: false,
+			},
+		});
+	}
+
+	/**
+	 * Queue up a callback that will make a request when an
+	 * instance becomes idle.
+	 * @param item Callback to run when intance becomes available. Should accept the instance as the first and only param, and return a promise that resolves when the request is complete.
```
dmsnell, line 195: typo: `intance` -> `instance`

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+	 * @public
+	 */
+	async enqueue(item: request): Promise<any> {
```
dmsnell, line 198: if we document the type then we don't need to repeat ourselves here, and it can thread through the code to help out in other places.

```ts
/**
 * Callback for handling PHP requests.
 *
 * @param instance PHP instance for handling the request, including all
 *                 cookies, HTTP headers, request body, etc...
 * @returns Resolve when finished handling the request.
 */
type RequestHandler = (instance: PhpInstance) => Promise<void>;
```

note that I've used `Promise<void>` here to communicate that we're not passing information back through the promise; we're only indicating that the request handling is finished.

Screenshot 2023-12-28 at 9 58 47 PM

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+
+			this.running.add(nextInstance);
+			info.requests++;
+
+			let request;
+
+			try {
+				const newInstances = await spawn(this);
+
+				// ... but, if we've just spanwed a fresh
```
dmsnell, line 254: typo: `spanwed` -> `spawned`

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+			return notifier;
+		}
+
+		// Given an instance, create a new callback that will clean up
+		// after the instance processes a request, and optionally
+		// will also kick off the next request.
+		const onCompleted = (instance) => async () => {
+			this.running.delete(instance);
+
+			reap(this);
```
dmsnell, line 232: why do we iterate over all of the pool's instances to reap them when we're dealing with a request itself right here. could we not skip some computation and immediately reap the passed-in `instance` by checking here?

we also seem to be missing type information and I found a type annotation as part of a JS comment that would be more helpful as a real type: `// php => PoolInfo`

```ts
const onCompleted = (instance: PhpInstance) => async () => {
	this.running.delete(instance);

	const instanceInfo = this.instanceInfo.get(instance);
	if (instanceInfo?.requests > this.maxRequests) {
		instanceInfo.active = false;
		this.instanceInfo.delete(instance);
		this.reap(instance);
	}
	…
}
```

it seems that if we do this right here, in the place where requests finish, then we can skip the call to `reap` in `enque()` and then get rid of the `reap` function, since we won't expect to have any instances ready to reap that haven't already been reaped. if we have a different path for failed instances then we can check there too, or turn the internal class `reap` method to operate on a single instance.

dmsnell, line 232: while investigating the types here I added a generic type parameter to the `Pool` class and it helped considerably to provide types around the class, as well as got rid of the awkward `type PhpInstance = any;`

```ts
interface Instance {}

class Pool<PhpInstance extends Instance> {
	instanceInfo = new Map<PhpInstance, PoolInfo>();

	…
}
```

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+				const newInstances = await spawn(this);
+
+				// ... but, if we've just spanwed a fresh
+				// instance, use that one instead.
+				if (newInstances.size) {
+					for (const instance of newInstances) {
+						nextInstance = instance;
+						break;
+					}
+				}
```
dmsnell, line 261: given how we're using this, why not return an array of newly-spawned processes, or return the most-recently spawned process? that way we wouldn't have to do the jumping jacks here to get the last one. it looks like this is the only place we're relying on the value returned by `spawn`

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+	}
+}
+
+/**
+ * Spawns new instances if the pool is not full.
+ * Returns a list of new instances.
+ * @param pool the pool object to work on
+ * @private
+ */
+const spawn = async (pool: Pool) => {
```
dmsnell, line 34: the name of the function and the comment seem to disagree, as `spawn` seems like it would _spawn a single instance_ but this is more like `spawnPoolInstances` or `spawnAllInstances` or `preSpawnInstances`

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+				// ... but, if we've just spanwed a fresh
+				// instance, use that one instead.
+				if (newInstances.size) {
+					for (const instance of newInstances) {
+						nextInstance = instance;
+						break;
+					}
+				}
+
+				request = next(nextInstance);
```
dmsnell, line 263: this line makes me think that `nextHandler` or `freeHandler` or `idleHandler` would be a more helpful name for `next`

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+			} catch (error) {
+				// Grab the reject handler from the notfier
```
dmsnell, line 265: typo: `notfier` -> `notifier`

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+			// so we can re-use it...
+			let nextInstance = instance;
+
+			const next = this.backlog.shift();
+			const info = this.instanceInfo.get(nextInstance);
+
+			this.running.add(nextInstance);
+			info.requests++;
+
+			let request;
```
dmsnell, line 249: notice that at this point, if we don't rename the `request` type, we will have a type and a variable with the same name in the same scope. they occupy different namespaces in TypeScript, which means there's no clash, but that's needlessly confusing if someone is trying to figure out what's happening

packages/wp-now/src/pool.ts
dmsnell, line 254: why are we preferring a newly-spawned instance over the one that just finished? is there any advantage to this inside our WASM world?

I would have guessed that passing the next request to the existing worker would exhaust the worker faster and prevent needing to create additional workers, which would lead to a lower overall memory use.

this also leads me to wonder about the pre-allocation of workers and if that's necessary. if we have `maxJobs` (or any more suitable name) and we're going to eagerly pay the costs of allocating them, why not defer allocation and initialization until those workers are required: that is, until we have a request come in when there are no non-active workers.

would you see any problem in reaping workers as soon as they complete a request and are over the max-requests threshold, and then only spawning new workers once there are no available non-active workers in the pool?

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
...
+			} catch (error) {
+				// Grab the reject handler from the notfier
+				// promise and run it if there is an error.
+				const resolver = this.resolvers.get(next);
+				this.resolvers.delete(next);
+				resolver.reject(fatal(this, nextInstance, error));
+				return;
+			}
+
+			const completed = onCompleted(nextInstance);
```
dmsnell, line 273: this line has no explanation, but it's calling `onCompleted` from within `onCompleted`, and is a bit dizzying trying to follow the flow of data inside of it, and inside the recursion.

I think here a different model that examines one request at a time could potentially remove a fair amount of logic and simplify everything in this flow. I'm curious to see if you also see the value and opportunity there.

packages/wp-now/src/pool.ts
```diff
@@ -0,0 +1,322 @@
+type PhpInstance = any;
+
+type request = (instance: PhpInstance) => Promise<any>;
```
dmsnell, line 3: Let's leave our types capitalized: `request` -> `Request`

additionally, it would help if we had a proper type for an instance and for a request. they don't need to be here, but somewhere would be good. here is a start I tossed together.

```ts
import { HTTPMethod } from '@php-wasm/universal';

interface HttpRequest {
	url: string;
	headers: Record<string, string>;
	method: HTTPMethod;
	files: Record<string, File>;
	body: string;
}

interface BaseInstance {
	request(data: HttpRequest): Promise<void>;
}
```

it could be that these already exist somewhere and I didn't see them. they will help later on, in places like `start-server.ts`

dmsnell (changes requested):
thanks for your continued efforts on this @seanmorris. I thought it was closer to being ready, but in my review here I uncovered some things that I think do warrant some extra attention:

 - an architectural question about the approach we're taking for eagerly pre-spawning instances and using new instances vs. reusing existing ones.
 - several questions about names that don't fully match function or intent, and lots of documentation that's going 99% of the way to being useful but misses out because it's not following the JSdoc conventions.
 - missing types that could thread documentation and consistency through the multiple layers of the system that PHP instances and HTTP requests take through this. specifically we should have _some_ type that describes the `PhpInstance`. if it's different for Node and the web that's fine, we can add a type parameter to `class Pool<PhpInstance extends BaseInstance> { }` and then have the `NodePool` and default pool extend `BaseInstance` however they want, but it should be there on all the functions inside `Pool` that work with requests. the `request` type is similar, and should take one of these `PhpInstance` values as its input.

I'm trying to summarize the issues here, but please check on all the comments as there are other items in there that would be good to address.

adamziel:
For posterity – this would be useful in php-wasm core as the web version of Playground suffers from the same memory leak issue.

adamziel:
I'm exploring an alternative in Playground core here:

https://github.com/WordPress/wordpress-playground/pull/990

adamziel:
https://github.com/WordPress/wordpress-playground/pull/990 just landed and enables a drop-in solution to the memory leak problem.

Instead of this:

```ts
const php = await NodePHP.load("8.0");
```

You would now do this:

```ts
import { rotatePHPRuntime } from '@php-wasm/universal';

const loadRuntime = async () => await NodePHP.loadRuntime("8.0");

const php = new NodePHP(await loadRuntime(), {
	documentRoot: '/test-root',
});

await rotatePHPRuntime({
	php,
	recreateRuntime: loadRuntime,
	maxRequests: 400
});
```

And the PHP runtime would be magically cleaned up and recreated every 400 requests.

Adapting `wp-now` will require a follow-up PR. Would you be up for exploring that @sejas or @bgrgicak?

sejas:
Nice!
I can create the PR for wp-now.
Instead of adding a maxRequest, it would be super cool to set it to ' auto' and the php instance could track if it needs "clean" the memory and rotate the runtime.

@adamziel , I think we need to release a new version of `@php-wasm/node`

sejas:
I created the PR for wp-now: https://github.com/WordPress/playground-tools/pull/152
It just needs to be updated with the new version of `@php-wasm/node` that contains the new function `rotatePHPRuntime`.

Currently is on version 0.5.6 that was released 18 days ago, and the new function was introduced 4 days ago on https://github.com/WordPress/wordpress-playground/pull/990

https://www.npmjs.com/package/@php-wasm/node

Screenshot 2024-02-05 at 10 31 47

adamziel:
> Instead of adding a maxRequest, it would be super cool to set it to ' auto' and the php instance could track if it needs "clean" the memory and rotate the runtime.

Agreed, I started a new issue to track it at https://github.com/WordPress/wordpress-playground/issues/998

> @adamziel , I think we need to release a new version of @php-wasm/node

You're right! v0.6.0 release is in progress now

PR #111, opened 2023-09-20
wp-now: Add proxy support

eruizdechavez:
## What?

This pull requests adds support for `https_proxy` and `http_proxy` environment variables to `wp-now` when downloading WordPress and SQLite at `start` time.

## Why?

Company/enterprise internet connections ofter require proxy configurations. While most GUI applications will inherit these configurations automatically from the system preferences, CLI tools still require manual configuration. It is a common practice to use these well-known environment variables for this purpose: `https_proxy`, `HTTPS_PROXY`, `http_proxy`, `HTTP_PROXY`.

## How?

This pull request addresses the problem of not being able to use `wp-now` while connected behind an HTTP(S) proxy by adding the required Agent configuration. To simplify the implementation, I am using [hpagent](https://www.npmjs.com/package/hpagent) which is an abstraction of the http.Agent with proxy support.

## Testing Instructions

1. Check out the branch.
2. Configure a proxy by exporting the proper proxy configurations.
  ```sh
  export http_proxy=<http://your-http-proxy-url:port>
  export https_proxy=<http://your-https-proxy-url:port>
  ```
3. Start wp-now with `nx preview wp-now start`. Ensure it will download something by either requesting a WordPress version not download previously, or removing your entire $HOME/.wp-now folder.
4. `wp-now` should download its dependencies and start as expected.

### DIFF
--- a/packages/wp-now/src/download.ts
+++ b/packages/wp-now/src/download.ts
@@ -11,6 +11,45 @@ import getWpNowPath from './get-wp-now-path';
 import getWordpressVersionsPath from './get-wordpress-versions-path';
 import getSqlitePath from './get-sqlite-path';
 import getWpCliPath from './get-wp-cli-path';
+import { HttpProxyAgent, HttpsProxyAgent } from 'hpagent';
+
+function httpsGet(url: string, callback: Function) {
+	const isBehindHttpProxy =
+		(process.env.http_proxy && process.env.http_proxy !== '') ||
+		(process.env.HTTP_PROXY && process.env.HTTP_PROXY !== '');
+
+	const isBehindHttpsProxy =
+		(process.env.https_proxy && process.env.https_proxy !== '') ||
+		(process.env.HTTPS_PROXY && process.env.HTTPS_PROXY !== '');
+
+	let agent;
+
+	if (isBehindHttpProxy || isBehindHttpsProxy) {
+		const urlParts = new URL(url);
+
+		let Agent;
+
+		if (urlParts.protocol === 'https:') {
+			Agent = HttpsProxyAgent;
+		} else if (urlParts.protocol === 'http:') {
+			Agent = HttpProxyAgent;
+		}
+
+		let proxy;
+
+		if (isBehindHttpsProxy) {
+			proxy = process.env.https_proxy ?? process.env.HTTPS_PROXY;
+		} else if (isBehindHttpProxy) {
+			proxy = process.env.http_proxy ?? process.env.HTTP_PROXY;
+		}
+
+		if (Agent && proxy) {
+			agent = new Agent({ proxy });
+		}
+	}
+
+	https.get(url, { agent }, callback);
+}
 
 function getWordPressVersionUrl(version = DEFAULT_WORDPRESS_VERSION) {
 	if (!isValidWordPressVersion(version)) {
@@ -41,7 +80,7 @@ async function downloadFile({
 		}
 		fs.ensureDirSync(path.dirname(destinationFilePath));
 		const response = await new Promise<IncomingMessage>((resolve) =>
-			https.get(url, (response) => resolve(response))
+			httpsGet(url, (response) => resolve(response))
 		);
 		statusCode = response.statusCode;
 		if (response.statusCode !== 200) {
@@ -96,7 +135,7 @@ async function downloadFileAndUnzip({
 
 		output?.log(`Downloading ${itemName}...`);
 		const response = await new Promise<IncomingMessage>((resolve) =>
-			https.get(url, (response) => resolve(response))
+			httpsGet(url, (response) => resolve(response))
 		);
 		statusCode = response.statusCode;
 
--- a/packages/wp-now/src/download.ts
+++ b/packages/wp-now/src/download.ts
@@ -14,38 +14,19 @@ import { output } from './output';
 import { isValidWordPressVersion } from './wp-playground-wordpress';
 
 function httpsGet(url: string, callback: Function) {
-	const isBehindHttpProxy =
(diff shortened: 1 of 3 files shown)


packages/wp-now/src/download.ts
```diff
@@ -11,6 +11,45 @@ import getWpNowPath from './get-wp-now-path';
...
+			proxy = process.env.http_proxy ?? process.env.HTTP_PROXY;
+		}
+
+		if (Agent && proxy) {
+			agent = new Agent({ proxy });
+		}
+	}
+
+	https.get(url, { agent }, callback);
+}
```
sejas: @eruizdechavez , what do you think about simplifying this function a bit?
I can see some redundant checks around `isBehindHttpsProxy` and `proxy`.

Here is a suggestion:

```suggestion
function httpsGet(url: string, callback: Function) {
	const proxy =
		process.env.https_proxy ||
		process.env.HTTPS_PROXY ||
		process.env.http_proxy ||
		process.env.HTTP_PROXY;
	let agent;

	if (proxy) {
		const urlParts = new URL(url);
		const Agent =
			urlParts.protocol === 'https:' ? HttpsProxyAgent : HttpProxyAgent;
		agent = new Agent({ proxy });
	}

	https.get(url, { agent }, callback);
}
```

packages/wp-now/src/download.ts
```diff
@@ -11,6 +11,45 @@ import getWpNowPath from './get-wp-now-path';
 import getWordpressVersionsPath from './get-wordpress-versions-path';
 import getSqlitePath from './get-sqlite-path';
 import getWpCliPath from './get-wp-cli-path';
+import { HttpProxyAgent, HttpsProxyAgent } from 'hpagent';
```
sejas: Let's move this import to line `7`, before the relative path imports.

packages/wp-now/public/with-node-version.js
```diff
@@ -6,9 +6,7 @@ import fs from 'fs';
 // Set the minimum required/supported version of node here.
 
 // Check if `--blueprint=` is passed in proccess.argv
-const hasBlueprint = process.argv.some((arg) =>
-	arg.startsWith('--blueprint=')
-);
+const hasBlueprint = process.argv.some((arg) => arg.startsWith('--blueprint='));
```
sejas: Can we undo this formatting change?

packages/interactive-code-block/src/lib/components/php-runner/php-loader.ts
```diff
@@ -24,9 +21,7 @@ export class PHPLoader extends EventTarget {
 			/** @ts-ignore */
 			'../../php-worker.ts?url&worker'
 		);
-		const worker = await spawnPHPWorkerThread(
-			workerScriptUrl
-		);
+		const worker = await spawnPHPWorkerThread(workerScriptUrl);
```
sejas: This formatting change is touching another package (`interactive-code-block`). Let's undo all the changes in this file.

sejas (commented):
@eruizdechavez, thanks for your work and contributing to `wp-now` and WordPress Playground.

I tested this PR without any proxy and it works well. Awesome!

Let's avoid changing the code format on `packages/interactive-code-block/src/lib/components/php-runner/php-loader.ts`, because is a different package.
You can avoid running the pre-commit formatter by appending the variable `-n` at the end  of your `git commit -m "example comment" -n` command.

I also added a suggestion to refactor `httpsGet` function a bit. Let me know what do you think.

I still need to test it with a proxy.

packages/wp-now/src/download.ts
eruizdechavez: I like the idea of simplifying the function. Let me try a couple of things and commit the updated version.

eruizdechavez:
Thanks for the feedback @sejas! I am not used to work with monorepos so I just assumed the linter knew better. I'll work on those suggestions and send and update after testing it.

packages/wp-now/src/download.ts
eruizdechavez: Done.

packages/wp-now/src/download.ts
eruizdechavez: Used VS Code reorganize imports, which moved all packages to the top, alphabetically and then the relative imports also alphabetically.

packages/wp-now/public/with-node-version.js
eruizdechavez: Done.

packages/interactive-code-block/src/lib/components/php-runner/php-loader.ts
eruizdechavez: Done.

sejas:
@eruizdechavez, awesome. Thanks for applying the feedback so quickly. I'll setup a proxy and confirm it works as expected, probably tomorrow. Thanks for the work.

sejas:
I tested it with a local proxy (Charles).

Testing instructions I've followed:
- Enable a local proxy, for example [Charles](https://www.charlesproxy.com/).
- Avoid using the proxy by default in your preferences > wifi
- Remove the latest WordPress version folder from `~/.wp-now/wordpress-versions/latest`
- Run:
```
nvm use
npm install
npx nx build wp-now
http_proxy=http://127.0.0.1:8888 node dist/packages/wp-now/cli.js start
```
- Observe the zip is downloaded and the Playground starts as expected.

## Screencast

https://github.com/WordPress/playground-tools/assets/779993/a8e74ecd-b080-4e6f-b361-1d0728a911d5



Thank you @eruizdechavez! I'll comment here after we deploy it to NPM, probably next week.

eruizdechavez:
👏 🎉 

Thanks!

PR #124, opened 2023-11-16
Add WordPress Playground plugin

seanmorris:
## What?

This adds the `collector` plugin for plugin-preview functionality and renames the plugin to WordPress Playground.

Requires:
https://github.com/WordPress/wordpress-playground/pull/745 .
https://github.com/WordPress/blueprints/issues/49

## Why?

Users want to be able to preview plugins on their own sites before actually committing to installing them live.

## Testing Instructions

- Checkout the project and copy the plugin to your local WordPress install:
```bash
git checkout sm-collector-plugin
cp packages/collector/ [PATH_TO_YOUR_WORDPRESS]/wp-content/plugins/
```
- Open your local WordPress install and activate the WordPress Playground plugin
- Start Playground by clicking on Tools > Sandbox Site in wp-admin
- After the site loads it should have the same content as your site (files and database)

### Testing Previews
- Open the file at `[PATH_TO_YOUR_WORDPRESS]/wp-content/plugins/collector/Collector.php`. Replace line 81 in line `Collector.php` with:
```php
const blueprintUrl = (query.get('blueprintUrl') ?? '').replace('https://wordpress.org', 'http://localhost:8010/proxy');
```
- Start the CORS proxy:
```bash
npm install -g local-cors-proxy
npx lcp --proxyUrl https://wordpress.org/
```
- Navigate to `plugins` > `add new`
- Search for "Akismet"
- You should now see a `Preview Now` button next to the `Install Now` button.
- Click `Preview now`, and you should see Playground boot your site with the plugin activated.


packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+				]);
+			}
+
+			steps.push(...[
+				{
+					step: 'login',
+					username: username,
+					password: fakepass,
+				}
+			]);
```
adamziel: That's a one-element array so this could be just:

```suggestion
			steps.push({
				step: 'login',
				username: username,
				password: fakepass,
			});
```

packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+				},
+				{
+					step: 'rm',
+					path: '/wordpress/wp-content/mu-plugins/1-show-admin-credentials-on-wp-login.php',
+				},
+			];
+
+			if(pluginUrl && pluginName)
+			{
+				steps.push(...[
```
adamziel: Neat trick! At the same time, it took me a good few seconds to go from "wait, what?", "oh, push accepts multiple arguments so this code is pushing many items, so effectively concatenating two arrays". While this code is more verbose, it could save future readers the same mental process:

```js
steps.push({
     step: 'writeFile',
     path: '/plugin.zip',
     // ...
});

steps.push({
     step: 'unzip',
     zipPath: '/plugin.zip',
     // ...
});

// ...etc
```

adamziel: Ah, I just realized this would work in the same way by just removing the `...[` part:

```js

steps.push(
    {
          step: 'writeFile',
          path: '/plugin.zip',
          // ...
    },
    {
          step: 'unzip',
          zipPath: '/plugin.zip',
          // ...
    },
    // ...
)
```

packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+					password: fakepass,
+				}
+			]);
+
+			const client = await startPlaygroundWeb({
+				iframe: document.getElementById('wp-playground'),
+				remoteUrl: `https://playground.wordpress.net/remote.html`,
+				blueprint: {
+					preferredVersions: {
+						wp: <?=json_encode(COLLECTOR_WP_VERSION);?>,
```
adamziel: I just checked and this actually works with `short_open_tag=0`. PHP is weird!

packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+						php: <?=json_encode(COLLECTOR_PHP_VERSION);?>,
+					},
+					phpExtensionBundles: ['kitchen-sink'],
+					steps,
+				},
+			});
+
+			client.isReady().then(() => client.goTo('/wp-admin/plugins.php'));
```
adamziel: This is an `async` function so you could do `await client.isReady()`

packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+		</div>
+		<div id = "wp-playground-main-area">
+			<iframe id = "wp-playground"></iframe>
+		</div>
+	</div>
+	<script type = "text/javascript">
+		const frame  = document.getElementById('wp-playground');
+		const zipUrl = <?=json_encode(COLLECTOR_DOWNLOAD_PATH);?>;
+
+		const username   = <?=json_encode(wp_get_current_user()->user_login);?>;
```
adamziel: That's a weird scenario, but is there an XSS here if `user_login` would for whatever reason contain the following string?
`</script><script>alert("I escaped");`

packages/collector/Collector_Content.php
```diff
@@ -0,0 +1,46 @@
+<?php
+function collector_iterate_directory($path, $prefix, $callback)
+{
+	$handle = opendir($path);
+
+	while($entry = readdir($handle))
+	{
+		if($entry === '.' || $entry === '..' || substr($entry, 0, 4) === '.git')
```
adamziel: Or .svn, or .mercurial etc. I wonder there's any blanket rule we could apply here, like "don't ship hidden directories" 🤔

packages/collector/Collector_Content.php
```diff
@@ -0,0 +1,46 @@
...
+function collector_export_tree()
+{
+	$flat = [];
+
+	$callback = function ($realPath, $packPath) use (&$flat, &$callback) {
+		if(is_file($realPath))
+		{
+			$flat[$packPath] = (object)[
+				'realPath' => $realPath,
+				'mode' => sprintf('%o', fileperms($realPath)),
```
adamziel: nice! I'm not sure if Emscripten supports file permissions, but it should still be handy in `wp-now` eventually

packages/collector/Collector_Content.php
```diff
@@ -0,0 +1,46 @@
...
+			];
+		}
+		else if(is_dir($realPath))
+		{
+			$flat[$packPath] = (object)[
+				'realPath' => $realPath,
+				'mode' => '040000',
+			];
+
+			collector_iterate_directory($realPath, '/wordpress', $callback);
```
adamziel: We may be in any directory at all, e.g. `/home/users/sean/public_html/`

adamziel: Ah, but that's fine because that's the **target** prefix, not the source directory. cool

packages/collector/Collector_Content.php
```diff
@@ -0,0 +1,46 @@
...
+		}
+	};
+
+	collector_iterate_directory('/wordpress/wp-content/', '/wordpress', $callback);
```
adamziel: We may be in any directory at all, e.g. `/home/users/sean/public_html/`.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,90 @@
...
+		{
+			if($table === 'wp_users' && (int) $record['ID'] === (int) wp_get_current_user()->ID)
+			{
+				$record['user_pass'] = wp_hash_password(collector_use_fakepass());
+			}
+
+			$insert = sprintf(
+				'INSERT INTO `%s` (%s) VALUES (%s);',
+				mysqli_real_escape_string($mysqli, $table),
+				implode(', ', array_map(fn($f) => "`" . mysqli_real_escape_string($mysqli, $f) . "`", array_keys($record))),
```
adamziel: Let's use `$wpdb` instead of `mysqli_*` functions whenever possible. Ideally there won't be a single `mysqli_*` call at all. WordPress may use different MySQL PHP extensions so there's no guarantee this function will be available.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,90 @@
...
+
+	return $tmpFiles;
+}
+
+function collector_get_db()
+{
+	static $mysqli;
+	if(!$mysqli)
+	{
+		$mysqli = mysqli_connect(DB_HOST, DB_USER, DB_PASSWORD);
```
adamziel: You should be able to reuse `global $wpdb` instead of starting a new database connection

packages/collector/Collector_Helpers.php
```diff
@@ -0,0 +1,28 @@
+<?php
+function collector_get_tmpfile($name, $type)
+{
+    $tmpName = tempnam('/tmp/', 'clctr-'. date('Y-m-d_H-i-s-') . $name . '-');
```
adamziel: Let's avoid writing potentially sensitive data to `/tmp` and store everything in memory instead

packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+Author: Sean Morris
+Version: 0.0.0
+Author URI: https://github.com/seanmorris/
+*/
+
+const COLLECTOR_DOWNLOAD_PATH   = '/wp-admin/?page=collector_download_package';
+const COLLECTOR_FINAL_ZIP       = '/tmp/collector-package.zip';
+
+define('COLLECTOR_PLAYGROUND_PACKAGE', ($_SERVER['SERVER_NAME'] === 'localhost')
+	? 'http://localhost:8081/index.js'
```
adamziel: once the runSql step lands, this line should no longer be necessary

packages/collector/Collector.php
```diff
@@ -0,0 +1,229 @@
...
+	: 'https://unpkg.com/@wp-playground/client/index.js'
```
adamziel: I feel uneasy about shipping an unpkg url here. I'd rather do one of these:

* Ship that file on playground.wordpress.net and use that URL here. I'm not sure if that will pass the plugin review, but it should – there are many plugins relying on external scripts.
* Ship `client/index.js` with this plugin – although that will add some maintenance burden to keep releasing new plugin versions whenever the client library gets updated

packages/collector/Collector.php
dawidurbanski: It's because `short_open_tag` in PHP is `<?`.

`<?=` is not a short open tag, it's something else used mostly in templating.

packages/collector/Collector.php
seanmorris: https://3v4l.org/IcF7CA#veol

```php
<?php
echo json_encode('</script><script>alert("I escaped");');
```
```
Output for 5.2.0 - 5.2.17, 5.3.0 - 5.3.29, 5.4.0 - 5.4.45, 5.5.0 - 5.5.38, 5.6.0 - 5.6.40, 7.0.0 - 7.0.33, 7.1.0 - 7.1.33, 7.2.0 - 7.2.34, 7.3.0 - 7.3.33, 7.4.0 - 7.4.33, 8.0.0 - 8.0.30, 8.1.0 - 8.1.24, 8.2.0 - 8.2.11
"<\/script><script>alert(\"I escaped\");"
```

https://jsfiddle.net/t9geyz51/

image

image

packages/collector/Collector.php
seanmorris: This notation was the result of a bunch of block-level refactoring. I'll fix that.

packages/collector/Collector.php
seanmorris: I was so happy on March 1st, 2012: https://www.php.net/ChangeLog-5.php#5.4.0

packages/collector/Collector_Helpers.php
seanmorris: Looks like there is no way to add a file handle pointing to `php://memory` to a `ZipArchive`.

`ZipArchives` don't expose any methods that allow a filehandle to be added directly:

https://www.php.net/manual/en/class.ziparchive.php#ziparchive.synopsis

We can't re-open `php://memory` since each time we do that, it references a different "file":

https://php.net/manual/en/wrappers.php.php#example-1264 (example # 5)

My workaround is to use the wordpress-recommended `get_temp_dir()`:

https://developer.wordpress.org/reference/functions/get_temp_dir/

packages/collector/Collector.php
seanmorris: @adamziel How should we proceed here?

packages/collector/Collector.php
adamziel: I feel uneasy about the `\/` escaping as the only barrier here. Is there any escaping function WordPress ships that would also escape the `<` and `>` characters? Maybe `esc_html`?

packages/collector/Collector_Helpers.php
adamziel: How about reading from that in-memory file handle and adding it to the archive [addFromString](https://www.php.net/manual/en/ziparchive.addfromstring.php)?

packages/collector/Collector.php
adamziel: https://github.com/WordPress/wordpress-playground/issues/778

bgrgicak:
@seanmorris I tried testing your PR for a Woo prototype, but couldn't get it to work.
This is what I did: 
- In the Playground project checked out your [runSQL branch](https://github.com/WordPress/wordpress-playground/pull/745)
- Merged trunk into it to get HTTP requests to work
- Started a local Playground
- Installed the collector plugin from your branch (using Plugin upload in the admin UI)
- Plugin installs and activates successfully
- After installing it any page I try to open fails with `ERR_FAILED`

packages/collector/Collector.php
seanmorris: > I feel uneasy about the `\/` escaping as the only barrier here. Is there any escaping function WordPress ships that would also escape the `<` and `>` characters? Maybe `esc_html`?

This is something I've thought about for a long time. If you're interpolating directly into JS, I don't think there is a stronger escaping function than json_encode. If you can break out of that, we need to open a bug report on PHP.net because JSON by definition should be valid Javascript.

I worry that adding extra measures around it could unintentionally weaken this in a non obvious way.

How would you feel about casting the input to JSON_encode to `(string)`? That should guarantee that the output of `json_encode` would be a valid string sequence in Javascript, rather than an integer, float, array or object.

packages/collector/Collector.php
seanmorris: @adamziel THANK YOU!!!!!!!!

packages/collector/Collector.php
adamziel: > JSON by definition should be valid Javascript.

The caveat here is that we're in a HTML document and need to worry about more escape sequences than just the JavaScript ones. That's what `esc_html` takes care of. `"</script>"` is a valid JavaScript but it would still terminate the script tag – and I know that case is covered, but there might be more. Relying on `esc_html` would provide an additional layer of safety.

adamziel: Also, I'm curious what's some prior art on this in other WordPress plugins where this kind of data passing is happening – there might be an established pattern this plugin could follow.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,75 @@
...
+
+	// Process in reverse order so wp_users comes before wp_options
+	// meaning the fakepass will be cleared before transients are
+	// dumped to the schema backup in the zip
+	foreach(array_reverse($tables) as $table)
+	{
+		file_put_contents($sqlFile, sprintf("-- %s\n", json_encode(['ACTION' => 'INSERT', 'TABLE' => $table])), FILE_APPEND);
+		$recordList = $mysqli->query(sprintf('SELECT * FROM `%s`', $mysqli->real_escape_string($table)));
+
+		while($record = $recordList->fetch_assoc())
```
seanmorris: @adamziel Is there a way to perform this pattern using `$wpdb`? I don't see a way to fetch records from a query sequentially. I've only found a way to load the entire result set into memory, or a single row and that's it. Is there a `fetch`-like method that I'm missing? I checked the docs, but didn't see one.

Unless `$wpdb` exposes this functionality, then I'd either have to run the same query multiple times, or load the entire result set into memory at once, neither of which is an optimal solution.

dawidurbanski: @seanmorris Maybe this is not greatly documented, but using for example `$wpdb->get_results("query")`, under the hood, inside the class, it calls `$this->query($query)` and this one later calls `$this->_do_query($query)`.

Looking at the source code of the last one, you can see that it sets the [`mysqli_result`](https://www.php.net/manual/en/class.mysqli-result.php) object (return value of [mysqli_query()](https://www.php.net/manual/en/mysqli.query.php)) in the `$wpdb->result` field:

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wpdb.php#L2348-L2350

The `mysqli_result` implements [`IteratorAggregate`](https://www.php.net/manual/en/class.iteratoraggregate.php) so you should be able to call `$wpdb->query("query")` directly (this doesn't create or return results array, opposite to `get_results()`, so no memory is being used), and then use `foreach` loop to traverse the buffer:

```php
global $wpdb;

$wpdb->query(
    $wpdb->prepare( "SELECT * FROM %1s", $wpdb->prefix . 'posts' )
);

foreach ( $wpdb->result as $row ) {
    var_dump( $row );
}
```

adamziel:
> @seanmorris I tried testing your PR for a Woo prototype, but couldn't get it to work. This is what I did:

@seanmorris please provide testing instructions in this PR's descriptions so that everyone can try it locally.

adamziel:
Let's not solve the generic case of Merging Blueprints for this one but ship a best guess of what that could look like and use the subsequent releases of this plugin to iterate on the merging concept.

packages/collector/Collector_Helpers.php
seanmorris: @adamziel That would force us to keep all the files in memory at once. Does `get_temp_dir()` have any security implications that I'm missing?

packages/collector/Collector_Db.php
seanmorris: @dawidurbanski I did notice that but it seems that the `$result` property is `protected`, and not actually part of the `public` api of the class. Can I expect that to remain constant as the class is updated and refactored in the future?

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wpdb.php#L154

dawidurbanski:
```
@since 0.71
```

Looks like it's been untouched since WordPress `0.71`. I would bet all my money on it being stable :D

dawidurbanski: > Can I expect that to remain constant as the class is updated and refactored in the future?

If that's about to happen, half of the internet is going down :)

seanmorris: One thing I'd be concerned about is if the underlying datatype were to change, but I did see the old file where the line originated, and it HAS been around a while, so I'll go with it.

packages/collector/Collector_Helpers.php
adamziel: > That would force us to keep all the files in memory at once

I agree that's not perfect. On the other hand, that's exactly what the user's browser where Playground is running will do so it's not a deal-breaker.

> Does get_temp_dir() have any security implications that I'm missing?

I'm worried that someone could read it from that temporary directory. As in, I'm not convinced that `get_temp_dir()` implies privacy. 

Here's just a few questions to answer there:

* What happens on shared hosting with poorly configured `/tmp` permissions?
* Is it possible that the temporary directory ever gets created inside `public_html`?
* How many developers set `WP_TEMP_DIR` to something public or readable by others?
* What if the PHP script errors out after the zip file is created and it lingers there for a long time?
* ...and there's more

OR the plugin could just store the data in memory for now and only revisit these questions once the memory isn't enough.

packages/collector/Collector_Db.php
seanmorris: https://github.com/WordPress/playground-tools/blob/c8e6a99f5ca0e8df43d1d2c5cfe7e5ba49ae9350/packages/collector/Collector_Db.php#L39-L59

packages/collector/Collector_Helpers.php
seanmorris: @adamziel How can we guarantee that we won't hit the memory limit in that case?

adamziel: @seanmorris We can't, but maybe that's fine. We can't guarantee we won't run out of disk space, either. Let's minimize the security risk by putting the sensitive data in memory and then find out what are the actual constraints before refining the approach.

Perhaps folks will be hesitant even about this memory-based approach and we'll have to reinvent the entire idea. Or perhaps, later on, we won't actually need to ship all the files in that zip? In a later version, the media files could be downloaded separately. Similarly, some plugins could eventually be transmitted as plugin directory URLs for Playground to download separately. And so on.

adamziel:
@seanmorris client.js aside, is this one ready for a re-review? I see there are some unresolved comments but the code also changed so I wonder if these comments are maybe addressed?

packages/collector/Collector.php
seanmorris: @adamziel 

https://github.com/WordPress/playground-tools/pull/124/files#diff-72b9ba31c538d25a3db28fcc6b6f58ca9b1fca1f66bab3947d565eaf85bbff07R143

packages/collector/Collector.php
seanmorris: @adamziel 

https://github.com/WordPress/playground-tools/pull/124/files#diff-72b9ba31c538d25a3db28fcc6b6f58ca9b1fca1f66bab3947d565eaf85bbff07R117

packages/collector/Collector.php
seanmorris: @adamziel

https://github.com/WordPress/playground-tools/blob/c8e6a99f5ca0e8df43d1d2c5cfe7e5ba49ae9350/packages/collector/Collector.php#L132

packages/collector/Collector.php
seanmorris: @adamziel 

https://github.com/WordPress/playground-tools/pull/124/files#diff-72b9ba31c538d25a3db28fcc6b6f58ca9b1fca1f66bab3947d565eaf85bbff07R76

packages/collector/Collector_Content.php
seanmorris: @adamziel How does this look for now? We can iterate on it in the future if the need presents itself:

https://github.com/WordPress/playground-tools/blob/c8e6a99f5ca0e8df43d1d2c5cfe7e5ba49ae9350/packages/collector/Collector_Content.php#L8

packages/collector/Collector_Content.php
seanmorris: @adamziel This has been refactored out.

packages/collector/Collector_Db.php
seanmorris: @adamziel https://github.com/WordPress/playground-tools/pull/124/files#diff-a9e52be37ae14ddda1c496c7a7f66a2139441a47bf645aab6ebb0e798cdb2ba3R46

packages/collector/Collector_Db.php
seanmorris: @adamziel https://github.com/WordPress/playground-tools/pull/124/files#diff-a9e52be37ae14ddda1c496c7a7f66a2139441a47bf645aab6ebb0e798cdb2ba3R21

packages/collector/Collector.php
adamziel: @seanmorris I'm still interested in what you may have found about this:

> I'm curious what's some prior art on this in other WordPress plugins where this kind of data passing is happening – there might be an established pattern this plugin could follow.

seanmorris: @adamziel It looks like the WP functions have a lot of moving parts where things could go wrong. How does this look?

https://github.com/WordPress/playground-tools/pull/124/files#diff-72b9ba31c538d25a3db28fcc6b6f58ca9b1fca1f66bab3947d565eaf85bbff07R79

packages/collector/Collector.php
adamziel: Hint: This link points to the `files` tab and it's showing this:

CleanShot 2023-12-11 at 23 03 36@2x

To share a specific change in a way that can be previewed even after the PR evolves, you typically want to link to a particular commit.

packages/collector/Collector.php
adamziel: Sure, `htmlspecialchars` seems tempting here. At the same time, some WordPress functions exist specifically because of deficiencies of these escaping functions provided by PHP. Let's pause additional code changes and discuss this for a while. What do some of the largest existing WordPress plugins do in this particular situation? How can that inform the escaping strategy in this place and in all similar places across the board?

packages/collector/Collector_Content.php
adamziel: Why not solve it from the get go? We know these hidden directories will be present in people's dev setups and, if it's not handled, developers will file issues about the slowness of the plugin. I wonder what strategy for filtering out these hidden directories would you use here?

Also, you seem to have filtered out the specific examples I used to illustrate the point. However, `.mercurial` is not an actual thing. Mercurial uses a `.hg` directory.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,82 @@
...
+		{
+			if($table === 'wp_users' && (int) $record['ID'] === (int) wp_get_current_user()->ID)
+			{
+				$record['user_pass'] = wp_hash_password(collector_use_fakepass());
+			}
+
+			$insert = sprintf(
+				'INSERT INTO `%s` (%s) VALUES (%s);',
+				esc_sql($table),
+				implode(', ', array_map(fn($f) => "`" . esc_sql($f) . "`", array_keys($record))),
```
adamziel: Does `esc_sql` escape backticks?

seanmorris: TIL you can have backticks in an sql identifier. Turns out you escape them by doubling, not with backslashes.

https://github.com/WordPress/playground-tools/blob/097ee8eb4811cd6594d030cbcee30530ad8992c0/packages/collector/Collector_Db.php#L54

https://github.com/WordPress/playground-tools/blob/097ee8eb4811cd6594d030cbcee30530ad8992c0/packages/collector/Collector_Db.php#L3-L5

packages/collector/Collector.php
seanmorris: @adamziel While I strongly agree that finding a solid, standard way of doing this is important, I think its exceedingly unlikely that someone would have XSS code in their administrative username. Would you agree that we can ship this as it stands, and iterate toward a more perfect solution in the future?

packages/collector/Collector_Content.php
seanmorris: @adamziel I think we can add a configuration option for this stuff. Is this something we can iterate on?

packages/collector/Collector.php
```diff
@@ -0,0 +1,219 @@
...
+/*
+Plugin Name: Collector
+Plugin URI: https://github.com/seanmorris/collector
+Description: Packages your WordPress install and sends it to Playground.
+Author: Sean Morris
+Version: 0.0.0
+Author URI: https://github.com/seanmorris/
+*/
+
+const COLLECTOR_DOWNLOAD_PATH = '/wp-admin/?page=collector_download_package';
```
bgrgicak: We should wrap this in `admin_url`.

packages/collector/Collector.php
```diff
@@ -0,0 +1,219 @@
...
+		exit();
+	}
+}
+
+function collector_plugin_menu()
+{
+	add_submenu_page(
+		NULL,
+		'Collector',
+		'Collector',
```
bgrgicak: Translations are missing.

packages/collector/Collector.php
```diff
@@ -0,0 +1,219 @@
...
+		'collector_render_playground_page',
+		NULL
+	);
+}
+
+function collector_render_playground_page()
+{?>
+	<div id = "wp-playground-wrapper">
+		<div id = "wp-playground-toolbar">
+			NOW WORKING INSIDE WORDPRESS PLAYGROUND &nbsp; [<a href = "/wp-admin" id = "goBack">RETURN</a>]
```
bgrgicak: Translations

bgrgicak: Update `admin_url`

packages/collector/Collector.php
```diff
@@ -0,0 +1,219 @@
...
+		'manage_options',
+		'collector_render_playground_page',
+		'collector_render_playground_page',
+		NULL
+	);
+}
+
+function collector_render_playground_page()
+{?>
+	<div id = "wp-playground-wrapper">
```
bgrgicak: Not a blocker, but we could move this to a template and keep the plugin code cleaner.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+function collector_esc_sql_identifier($tableName) {
+	return str_replace('`', '``', esc_sql($tableName));
+}
+
+function collector_dump_db($zip)
+{
+	$tables   = collector_get_db_tables();
```
bgrgicak: We should at least add a filter here to allow excluding tables.

bgrgicak (commented):
I added some comments for myself to work on. The plugin looks good and these are just details.

@adamziel I'm still not sure if we should just dump the entire DB. It seems extremely risky. I would by default add some extra protection like not syncing users and having support for excluding some option rows. It would be nice if we could detect if an option looks like a password and exclude it, but I'm not sure how would we do that.

adamziel:
> @adamziel I'm still not sure if we should just dump the entire DB. It seems extremely risky. I would by default add some extra protection like not syncing users

Not syncing passwords for sure. About user accounts themselves, we'd lose post authors and probably other data relationships. I also wonder how often folks will want to preview the site from another user's perspective. Maybe that could be a checkbox somewhere, or a select box "[ ] Import other user accounts ((i) learn about the risks)" 🤔 

> and having support for excluding some option rows. It would be nice

that would be nice for sure

> if we could detect if an option looks like a password and exclude it, but I'm not sure how would we do that.

We probably can't. With enough work, we could detect sha1, md5 etc. in serialized data, but that's still not enough information to say whether that's a password or, say, cache key.

bgrgicak:
We could make it optional to import user data and some options with a settings page (in another PR).

For users, I would replace PII data with fake data, and maybe not import metadata.

For options, we could have a blacklist for option names. For example, if the name contains _secret_ or _key_.

Ideally plugin authors could add a blacklist to blueprint.json.

marcarmengou:
I leave this here in case it can be useful to you in some way. The Plugin exploration is one of the tools of the Toys for Playground plugin:

https://playground.wordpress.net/?plugin=toys-for-playground&networking=yes&url=/wp-admin/plugin-install.php&mode=seamless

You will see that when browsing a theme or a plugin does not load, that is due to problems with the Playground `scopes`. But you will not have any problem when using the plugin locally or on a live site.

Here is the plugin if you want to test it locally or on a live site: https://wordpress.org/plugins/toys-for-playground/

bgrgicak:
Hey @marcarmengou!

Thanks for sharing Toys for Playground with us, it's a great example of what Playground can enable.

> You will see that when browsing a theme or a plugin does not load, that is due to problems with the Playground scopes.

Could you please explain this a bit more and provide some examples of the errors you see in [a new issue](https://github.com/WordPress/wordpress-playground/issues/new)?

packages/collector/Collector.php
bgrgicak: Updated in 1ee4c4e97e29015cbc5f8e3f751012dd69c76344 to [use WordPress escaping.](https://developer.wordpress.org/apis/security/escaping/)

packages/collector/Collector_Content.php
bgrgicak: ec5b36741505c08173e5772ad4ba52a6b75f75e8 adds support for skipping all hidden files and folders. We can add a filter in the future to allow for more granular importing.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,90 @@
...
+
+	fwrite($sqlFileHandle, sprintf("-- %s\n", json_encode(['SECTION END' => 'RECORDS'])));
+
+	fseek($sqlFileHandle, 0);
+	$content = '';
+	while (!feof($sqlFileHandle)) {
+		$content .= fread($sqlFileHandle, 8192);
+	}
+	$zip->addFromString('schema/_Schema.sql', $content);
+	fclose($sqlFileHandle);
```
bgrgicak: @adamziel this is how I was able to write the SQL dump in memory and add it to the zip. 
But it feels like an overkill. Should I just write it to the `$content` variable immediately instead of messing with  `php://memory`?

bgrgicak: I swapped `php://memory` with a variable. Let me know if you see any benefits in using `php://memory`.

packages/collector/Collector.php
bgrgicak: Leaving this one for a future PR

packages/collector/Collector_Db.php
bgrgicak: We can do this in a future pr

bgrgicak (commented):
This should be ready for another review. The PR size grew a lot because I had to add `ZipStream`.

packages/collector/Collector_Helpers.php
bgrgicak: Switched to streaming in 22b6cd46f73f793a980c37a4801523cb686371fd

packages/collector/Collector.php
```diff
@@ -0,0 +1,250 @@
...
+ * @package Collector
+ * @version 0.0.0
+ */
+/*
+Plugin Name: Collector
+Plugin URI: https://github.com/seanmorris/collector
+Description: Packages your WordPress install and sends it to Playground.
+Author: Sean Morris
+Version: 0.0.0
+Author URI: https://github.com/seanmorris/
```
adamziel: As this plugin is the exploratory ground for a canonical plugin and then merging in core, let's go with Author: WordPress Contributors to communicate it's a collective effort and invite new folks to participate. I'll be adjusting the same thing in the Playground Block plugin also hosted in this repo.

packages/collector/Collector.php
```diff
@@ -0,0 +1,250 @@
...
+}
+
+function collector_plugins_loaded()
+{
+	if(!current_user_can(ADMIN_PAGE_CAPABILITY))
+	{
+		return;
+	}
+
+	if(home_url($_SERVER['REQUEST_URI']) === get_collector_admin_page_url())
```
adamziel: Is this how plugins typically route requests? Something seems weird about that but WordPress surprised me more than once.

packages/collector/Collector.php
```diff
@@ -0,0 +1,250 @@
...
+				},
+				{
+					step: 'unzip',
+					zipPath: '/data.zip',
+					extractToPath: '/wordpress',
+				},
+				{
+					step: 'rm',
+					path: '/data.zip',
+				},
```
adamziel: This should work, too:

```suggestion
				{
					step: 'unzip',
					zipFile: {
						'resource': 'url',
						'url': zipUrl,
					},
					extractToPath: '/wordpress',
				},
```

Ideally I'd like to introduce a syntax that would enable something like:

```ts
				{
					step: 'runSql',
					sql: "https+zip://mysite.com/sql-dump.zip"
				},
```

packages/collector/Collector.php
```diff
@@ -0,0 +1,250 @@
...
+			];
+
+			blueprint.preferredVersions = {
+				wp: '<?php echo esc_js(COLLECTOR_WP_VERSION); ?>',
+				php: '<?php echo esc_js(COLLECTOR_PHP_VERSION); ?>',
+			};
+
+			const client = await startPlaygroundWeb({
+				iframe: document.getElementById('wp-playground'),
+				remoteUrl: `https://playground.wordpress.net/remote.html`,
```
adamziel: This is fine for V1, but eventually let's put base URL in a config file and enable local overrides for working with a local Playground dev server.

packages/collector/Collector.php
bgrgicak: No, they typically use API endpoints or at least register admin pages.
I didn't want to modify this part to much in this PR, but we should address it in the future to avoid our custom security checks and just rely on WP.

packages/collector/Collector.php
```diff
@@ -0,0 +1,250 @@
...
+				blueprint
+			});
+
+			await client.isReady();
+
+			client.goTo('/wp-admin/plugins.php');
+		})();
+
+		const goBack = document.getElementById('goBack');
+		const goBackClicked = event => {
```
adamziel: What is the intention behind this function? I'm struggling a bit 😅

packages/collector/Collector.php
```diff
@@ -0,0 +1,250 @@
...
+				)
+			),
+			esc_attr( $plugin['name'] ),
+			__( 'Preview Now', TRANSLATE_DOMAIN )
+		);
+
+		array_unshift($action_links, $preview_button);
+
+		// Use only a single Blueprint json for now:
+		break;
```
adamziel: Sounds like we don't need a loop here?

packages/collector/Collector.php
bgrgicak: Nothing, I removed it now.

packages/collector/Collector_Content.php
```diff
@@ -0,0 +1,18 @@
...
+	$handle = opendir($path);
+
+	while($entry = readdir($handle))
+	{
+		if($entry === '.' || $entry === '..' || substr($entry, 0, 1) === '.')
+		{
+			continue;
+		}
+
+		$realPath = realpath($path . '/' . $entry);
```
adamziel: would https://developer.wordpress.org/reference/functions/path_join/ be a better fit here?

packages/collector/Collector_Content.php
```diff
@@ -0,0 +1,18 @@
...
+		$packPath = substr($realPath, strlen($prefix));
+
+		$callback($realPath, $packPath);
```
adamziel: Could that be a generator that yields the computed paths instead of calling a callback? Callbacks in PHP seem like a bit less natural choice than in, say, JS. Happy to discuss.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,83 @@
...
+{
+	$tables   = collector_get_db_tables();
+	$sql_dump = '';
+
+	$sql_dump .= sprintf("-- %s\n", json_encode(['SECTION START' => 'SCHEMA']));
+
+	foreach($tables as $table)
+	{
+		$sql_dump .= sprintf("-- %s\n", json_encode(['ACTION' => 'DROP', 'TABLE' => $table]));
+		$sql_dump .= sprintf("DROP TABLE IF EXISTS `%s`;\n", collector_esc_sql_identifier($table));
```
adamziel: Is there any `$wpdb` escaping utility better suited for the job? I imagine this could fail in a few ways when backslashes are involved. We may not be able to rely on `mysql_` functions as they're not available in Playground, but perhaps WordPress provides something custom.

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,83 @@
...
+		$sql_dump .= sprintf("-- %s\n", json_encode(['ACTION' => 'CREATE', 'TABLE' => $table]));
+		$sql_dump .= preg_replace("/\s+/", " ", collector_dump_db_schema($table)) . "\n";
```
adamziel: Why replace the whitespace here? What if a default column value contains two spaces?

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	$sql_dump .= sprintf("-- %s\n", json_encode(['SECTION END' => 'SCHEMA']));
```
adamziel: Is this information used anywhere? Or could it be safely removed (or replaced with plaintext `-- end of schema`?)

packages/collector/Collector_Db.php
```diff
@@ -0,0 +1,83 @@
...
+	$sql_dump .= sprintf("-- %s\n", json_encode(['SECTION START' => 'RECORDS']));
+
+	global $wpdb;
+
+	// Process in reverse order so wp_users comes before wp_options
+	// meaning the fakepass will be cleared before transients are
```
adamziel: If the ordering is important here, let's make sure `$tables` are sorted. At the moment the seem to be no guarantees about what comes first. However, I'm not sure what clears the fakepass as the comment mentions, why is that important, and what is the relation with transients.

packages/collector/Collector_Helpers.php
```diff
@@ -0,0 +1,23 @@
...
+function collector_get_tmpfile($name, $type)
+{
+    return $name . '-package-'. date('Y-m-d_H-i-s') . '.' . $type;
+}
+
+function collector_get_fakepass()
+{
+    $fakepass = base64_encode(random_bytes(128));
+
+    set_transient('COLLECTOR_FAKE_PASSWORD', $fakepass, 60 * 5);
```
adamziel: What's the value of storing it and passing as transient vs hardcoding it as `"password"`?

packages/collector/Collector_Zip.php
```diff
@@ -0,0 +1,44 @@
...
+		{
+			$zip->addFileFromPath($packPath, $realPath);
+		}
+		else if(is_dir($realPath))
+		{
+			collector_iterate_directory($realPath, ABSPATH, $callback);
+		}
+	};
+
+	collector_iterate_directory(ABSPATH . '/wp-content/', ABSPATH, $callback);
```
adamziel: Would `RecursiveDirectoryIterator` solve the same problem without a custom callback-based utility?

packages/collector/Collector_Zip.php
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function collector_open_zip()
+{
+	$options = new Archive();
+	$options->setSendHttpHeaders(true);
+	$zip = new ZipStream(
```
adamziel: Nice!

packages/collector/Collector_Zip.php
```diff
@@ -0,0 +1,44 @@
...
+		collector_get_tmpfile('collector', 'zip'),
```
adamziel: `collector_get_tmpfile` is only used here. What do you think about inlining it?

```suggestion
		'collector-package-'. date('Y-m-d_H-i-s') . '.zip',
```

or even just

```suggestion
		'export.zip',
```

adamziel:
Thanks @bgrgicak ! I left some notes. A more general note would be to rename this to, say, "WordPress Zip Export" plugin. This way it would be all about `WordPress -> Portable ZIP Export` and become a natural fit for exploring the [Site Transfer Protocol](https://core.trac.wordpress.org/ticket/60375) and the WordPress to WordPress migration part of the [Data Liberation](https://github.com/WordPress/data-liberation/discussions/53) proposal.

cc @StevenDufresne @tellyworth @jordesign

packages/collector/Collector_Db.php
adamziel: We'll have to stream above a certain database size, but for now a variable sounds great. Streaming will come with its own set of problems to resolve like, pausing, resuming, seeking, dealing with deltas etc.

bgrgicak:
@adamziel  I agree that we should rename the plugin, but I'm not sure if "WordPress Zip Export" matches the functionality.
Today this plugin allows site previews, so a name like WordPress Preview, WordPress Snapshot, or WordPress Sandbox would be closer to its functionality.
I would like to understand what functionality we plan for this plugin to have. For example, do we want to enable previews (sandboxes) without installing something? Should we add a Download ZIP button to the Tools menu to allow site content downloads?

What would be the best place to discuss this name and who should we involve?

packages/collector/Collector_Helpers.php
bgrgicak: Not sure, removing this also

adamziel:
> @adamziel I agree that we should rename the plugin, but I'm not sure if "WordPress Zip Export" matches the functionality. Today this plugin allows site previews, so a name like WordPress Preview, WordPress Snapshot, or WordPress Sandbox would be closer to its functionality.

You're right! Howe about just WordPress Playground, then?

Later on let's split into two plugins: [Site Transfer](https://core.trac.wordpress.org/ticket/60375), and Playground features.

> I would like to understand what functionality we plan for this plugin to have.

To me, this is an exploratory ground for the [Site Transfer Protocol](https://core.trac.wordpress.org/ticket/60375). The rest of it, like embedded previews, installing plugins etc. is just a demo of that's enabled by the site transfer.

> do we want to enable previews (sandboxes) without installing something?

Yes, that would be useful.

> Should we add a Download ZIP button to the Tools menu to allow site content downloads?

Nice idea! Perhaps embedding the larger Playground settings modal would make sense here – that way you could also play with the WordPress version, PHP version, PHP extensions, and so on.

> What would be the best place to discuss this name and who should we involve?

It's not clear-cut, but:

* For user-facing features, like zip downloads or previews without installation, a new issue or discussion in this repo would be perfect.
* For site transfer-related topics, let's use the [Site Transfer Protocol](https://core.trac.wordpress.org/ticket/60375) issue.

These threads are intertwined and, occasionally, we'll get the place wrong. That's fine.

packages/collector/Collector.php
bgrgicak: I added filters to both Playground urls in d4cb48d53d1646b345fa62002bdd548854f4911b

packages/collector/Collector_Zip.php
bgrgicak: Yes, updated here a1464c95f95512c6b4ca1587e4d2e96c8371e9eb

packages/collector/Collector.php
bgrgicak: Turns out we can't use the rest API here because we need a zip URL. The rest API would require us to have a blueprint step where we can pass a file previously downloaded using `fetch`.

I updated the code to use the admin page to generate ZIPs which would be a more standard solution. 8adf17168f645d5751085adc8f9a19d32c68eba0

packages/collector/Collector_Db.php
bgrgicak: I updated this to use `esc_sql`

packages/collector/Collector_Db.php
bgrgicak: These are comments, I removed them all to keep the code simple.

packages/collector/Collector_Db.php
bgrgicak: I didn't find a reason why this would be needed, so I removed it.

packages/collector/Collector_Db.php
bgrgicak: The schema doesn't work if we don't replace the whitespace. I tried a few things with regex to limit it, but it all resulted in the dump not working.

I would like to merge this code as is and come back to it later if it becomes an issue.

packages/playground/playground.php
```diff
@@ -0,0 +1,160 @@
...
+		),
+		'playgroundRemoteUrl' => apply_filters(
+			'playground_playground_remote_url',
+			esc_url('https://playground.wordpress.net/remote.html'),
+		)
+	]);
+	wp_enqueue_script('playground');
+}
+
+function get_playground_download_page_url()
```
adamziel, line 59: nice!

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+defined('ABSPATH') || exit;
+
+function playground_escape_array($array)
+{
+	global $wpdb;
+	$escaped = array();
+	foreach ($array as $value) {
+		if (is_numeric($value))
```
adamziel: Nitpick: let's use braces consistently `{`

adamziel: More substantially – this seems like a much better way of escaping the data, thank you!

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
...
+{
+	global $wpdb;
+
+	$tables   = playground_get_db_tables();
+	$sql_dump = array();
+
+	foreach ($tables as $table) {
+		array_push(
+			$sql_dump,
+			sprintf("DROP TABLE IF EXISTS `%s`;", esc_sql($table)),
```
adamziel: What's the difference between `esc_sql` and `$wpdb->prepare()`? Would it make sense to stick to one of these?

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
...
+	foreach ($tables as $table) {
+		$records = $wpdb->get_results(
+			sprintf(
+				'SELECT * FROM `%s`',
+				esc_sql($table)
+			),
+			ARRAY_A
+		);
+
+		if ($wpdb->last_error) {
```
adamziel, line 46: Is this reset on each query? Or is it possible this would be populated here from the query before the `SELECT * FROM %s`?

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
...
+			error_log($wpdb->last_error);
+			continue;
+		}
+
+		foreach ($records as $record) {
+			array_push(
+				$sql_dump,
+				sprintf(
+					'INSERT INTO `%1$s` (%2$s) VALUES (%3$s);',
```
adamziel: Ditto the question from above – could this sprintf be replaced with `$wpdb->prepare()`?

packages/playground/src/playground-db.php
bgrgicak: This is reseted and it applies only to the current query.

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
...
+					esc_sql($table),
+					playground_escape_array(
+						array_keys($record)
+					),
+					playground_escape_array(array_values($record))
+				)
+			);
+		}
+	}
+	$zip->addFile('schema/_Schema.sql', implode("\n", $sql_dump));
```
adamziel, line 65: Nice! This will do for now, and in the future we'll find a way of streaming really large dumps, e.g. by zipping and sending over at most 1MB at a time

packages/playground/src/playground-db.php
bgrgicak: `$wpdb->prepare()` will add quotes around table name, and it's a practice in WP to not add table names with `$wpdb->prepare()`.

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
...
+
+}
+
+function playground_get_db_tables()
+{
+	global $wpdb;
+	$tables = $wpdb->get_results('SHOW TABLES', ARRAY_N);
+	return array_map(fn ($t) => $t[0], $tables);
```
adamziel: Arrow functions were introduced in PHP 7.4, let's make sure this plugin runs on the same PHP versions as WordPress core, that is >= 7.0. How easy would it be to add a phpcs check for that?

packages/playground/src/playground-db.php
bgrgicak: No, just like with table names `$wpdb->prepare()`  would wrap params in quotes.

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,87 @@
...
+		ARRAY_A
+	);
+	if ($wpdb->last_error) {
+		error_log($wpdb->last_error);
+		return '';
+	}
+	if (!isset($schema['Create Table'])) {
+		return '';
+	}
+	return preg_replace("/\s+/", " ", $schema['Create Table']);
```
adamziel: Won't this introduce unintended changes in statements like `CREATE TABLE test ( name varchar(255) default "FIRST      LAST" );`?

packages/playground/src/playground-db.php
bgrgicak: Done [2942c73](https://github.com/WordPress/playground-tools/pull/124/commits/2942c733df3e4d3cf8eef68862d689788ff8cbec)

packages/playground/src/playground-zip.php
```diff
@@ -0,0 +1,46 @@
...
+
+require __DIR__ . '/playground-db.php';
+
+use ZipStream\ZipStream;
+use ZipStream\Option\Archive;
+
+function playground_zip_wp_content($zip)
+{
+	$root_dir = WP_CONTENT_DIR;
+	$directory = new \RecursiveDirectoryIterator($root_dir, \FilesystemIterator::FOLLOW_SYMLINKS);
```
adamziel, line 13: Lovely!

packages/playground/src/playground-zip.php
```diff
@@ -0,0 +1,46 @@
...
+	$iterator = new \RecursiveIteratorIterator($directory);
+	$regex = new \RegexIterator($iterator, '/^.+\/(?!.*\.\.).+$/i', \RecursiveRegexIterator::GET_MATCH);
```
adamziel: Let's document what that regex is doing, I can't tell without stopping here for a few minutes

packages/playground/src/playground-db.php
adamziel: Aha, gotcha! Is `sprintf()` typically used in WordPress core in cases like that? If yes, we're good. If not, is it possible  do here what core does?

packages/playground/src/playground-db.php
adamziel: Is there anything wrong with quotes, though?

packages/playground/src/playground-db.php
bgrgicak: The query breaks the dump if we don't remove the extra whitespace.
I tried limiting the replace by doing something like this `preg_replace("/([,(`])\s\s+/", "$1 ", $schema['Create Table'])` but the dump didn't work. I probably missed something in my regex, but couldn't figure it out.

If this isn't a big deal, I would prefer to skip it.

adamziel: Aha, I see! That's because we assume a single query per line. But that means we only have to remove newlines, not all whitespace, right?

adamziel: I just ran a very naive check of whether removing newlines could mess up the output of CREATE TABLE, and it seems like we're good in basic cases:

```
create table newline_test ( id int, name varchar(255) default "a
                   b" );
insert into newline_test (id) VALUES (1);
show create table newline_test;
```

```
CREATE TABLE `newline_test` ( `id` int(11) DEFAULT NULL, `name` varchar(255) DEFAULT 'a\n b' ) ENGINE=InnoDB DEFAULT CHARSET=latin1
```

In a future release, we could use the MySQL tokenizer shipped with the `sqlite-database-integration` plugin to normalize these queries not as strings but as semantic data structures. Either here, on write, or in Playground, on read.

adamziel: For more context – the "single query per line" constraint is the limitation of the SQL query parser shipped with that plugin. It knows how to read a single query, but it doesn't yet have an API to communicate "I reached the finishing semicolon and now I'm starting to read another query". It can be implemented with relative ease, but it isn't yet.

packages/playground/src/playground-zip.php
bgrgicak: Done in  e0cb7c58fe6785d24fbc53006677c1d8f5667f76

packages/playground/src/playground-db.php
bgrgicak: Done in  e0cb7c58fe6785d24fbc53006677c1d8f5667f76

packages/playground/src/playground-db.php
bgrgicak: Yes, they break queries. 

Examples
- `SELECT * FROM 'wp_posts'`
- `INSERT INTO 'wp_options' ('1,"test","yes"')`

packages/playground/src/playground-db.php
bgrgicak: Sometimes it's simple 🙂 
That regex worked, I updated it in ee10c8ff57078bc6c55942257163a8fa27c821ee and added a comment.

packages/playground/src/playground-db.php
bgrgicak: You would usuall see `"SELECT * FROM $wpdb->posts"` but this would require us to save the escaped value in a variable, or contact strings `'DROP TABLE IF EXISTS `' . esc_sql($table) . '`;'` I think that `sprintf` is a ok solution here.

packages/playground/src/playground-db.php
adamziel: Ah, they're singular quotes, not double quotes. Gotcha!

packages/playground/src/playground-db.php
adamziel: The [doc page for esc_sql](https://developer.wordpress.org/reference/functions/esc_sql/) says this:

> Be careful in using this function correctly. It will only escape values to be used in strings in the query. That is, it only provides escaping for values that will be within quotes in the SQL (as in field = '{$escaped_value}'). If your value is not going to be within quotes, your code will still be vulnerable to SQL injection. For example, this is vulnerable, because the escaped value is not surrounded by quotes in the SQL query: ORDER BY {$escaped_value}. As such, this function does not escape unquoted numeric values, field names, or SQL keywords.

Here's how it handles unsanitized inputs:

```php
<?php
echo "`" . esc_sql("A backtick ` and a backslash " . chr(92)) . "`";
// `A backtick ` and a backslash \\`
```

Here's what `$wpdb->quote_identifier` does:

```php
<?php
echo $wpdb->quote_identifier("A backtick ` and a backslash " . chr(92))
// `A backtick `` and a backslash \`
```

The latter seems to be correct:

```sql
create table `te``st\` (id int);
show tables;
```

```
Tables_in_test
--
te`st\
```

packages/playground/src/playground-zip.php
adamziel: Aha, thank you! This will exclude the `.htaccess` files, the `.ht.sqlite` database file, and other potentially desirable files. I'm not sure what's the right thing to do here, but we could start by allowing those two.

packages/playground/src/playground-db.php
bgrgicak: In that case there is no point in escaping table names, which should be fine as they are coming from the DB.
Should I just do `"SELECT * FROM $table"`?

adamziel: Well, in the (highly unlikely) case I listed above, `"SELECT * FROM $table"` would not work, as it would become...

```sql
SELECT * FROM te`st\
```

However,  `sprintf("SELECT * FROM %s", $wpdb->quote_identifier($table))` would work as it would become...

```sql
SELECT * FROM `te``st\`
```

which is valid MySQL SQL.

packages/playground/src/playground-zip.php
bgrgicak: Are these files used by Playground in any way?
If not we may not actually need them now.

packages/playground/src/playground-db.php
bgrgicak: I see now. What about esc_sql?
It seems like it's not adding any security so we can remove it.

bgrgicak: Sorry about all the questions I would just like to ensure we are going in the same direction.

packages/playground/src/playground-zip.php
adamziel: `.htaccess` isn't and a `@TODO` note would be good enough for now. The `.ht.sqlite` file would normally be used if available, but in this case we're moving the data as SQL dump so I guess that isn't used either. A comment or an issue should do the trick.

packages/playground/src/playground-db.php
adamziel: Agreed about removing `esc_sql`. `sprintf("SELECT * FROM `%s`", esc_sql($table))` would return invalid SQL:

```
SELECT * FROM `te`st\\`
```

`$wpdb->quote_identifier` looks like a good replacement for `esc_sql`

packages/playground/src/playground-zip.php
bgrgicak: Done in c02934d36a2fe2dec4ad6d871179182c3a8f03e6

packages/playground/src/playground-db.php
bgrgicak: I updated this in c02934d36a2fe2dec4ad6d871179182c3a8f03e6.
Thank you for your help on this. It took a while, but I think that we landed on a solid implementation.

bgrgicak:
@adamziel I moved the code [to a PR on this repo](https://github.com/WordPress/playground-tools/pull/154).
Let's move the conversation there. All your current comments in this PR should have already been addressed.

PR #132, opened 2023-12-08
Adding debug flags

seanmorris:
## What?

This PR adds debugging flags to wp-now.

## Why?

This allows for more robust debugging of php-wasm on the command line.

## How?

By adding the following flags to wp-now:

* `inspect` - Connect to a Node debugging client.
* `inspect-brk` - Connect to a Node debugging client, break immediately on script execution start.
* `trace-exit` - Prints a stack trace whenever an environment is exited proactively, i.e. invoking process.exit().
* `trace-uncaught` - Print stack traces for uncaught exceptions; usually, the stack trace associated with the creation of an Error is printed, whereas this makes Node.js also print the stack trace associated with throwing the value (which does not need to be an Error instance).
* `trace-warnings` - Print stack traces for process warnings (including deprecations).

## Testing Instructions

Check out the branch

```bash
git checkout sm-debugging-flags
```

Run wp-now with `--trace-warnings`

```bash
npx wp-now start --trace-warnings
```

You should see a stack trace with any warnings printed:

```
> nx run wp-now:preview start --trace-warnings --path=/home/sean/mattic/playground-tools

(node:1708766) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("/home/sean/mattic/playground-tools/packages/nx-extensions/src/executors/built-script/loader.mjs", pathToFileURL("./"));'
    at createModuleLoader (node:internal/modules/esm/loader:534:17)
    at loadESM (node:internal/process/esm_loader:21:19)
    at runMainESM (node:internal/modules/run_main:53:21)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:5)
    at node:internal/main/run_main_module:23:47
Starting the server......

```

**Kill the above script with ctrl+c or switch to a new terminal to continue.**

Open Chrome and enable DevTools by pressing `F12`.

Run wp-now with the `--inspect-brk` flag:

```
npx wp-now start --inspect-brk
```
The NodeJS debug icon should appear in green at the top left of the DevTools window, click on it.

image

You should see the step-debugger open, with the script paused on the first line:

image


adamziel:
Thank you for providing that description! I haven't tested yet, but wanted to note these information would make a great addition to README.md – let's add them in there.

seanmorris:
> Thank you for providing that description! I haven't tested yet, but wanted to note these information would make a great addition to README.md – let's add them in there.

That's been added:

https://github.com/seanmorris/playground-tools/blob/sm-debugging-flags/README.md

https://github.com/WordPress/playground-tools/blob/6ddbfae25d0856ded44d6e0c76b9f7d9a10ffe19/README.md?plain=1#L29-L53

PR #137, opened 2024-01-07
wp-now load .zip downloaded from wordpress-playground

rhildred:
## What?

Bring .zip from wordpress-playground into node.js with wp-now. I also exposed the existing wp-cli code on the command line with `@wp-now/wp-now wp --version` for further testing.

## Why?

I am leading a class in wordpress this term. There is a lab each week, for which students hand in a .zip of their site. Wordpress-playground offers the ability for students to produce the .zip that I am looking for.  To mark the lab, I have code that makes a folder for each student. I want to be able to start a terminal in vscode in the student folder, type `npx @wp-now/wp-now start` and see the student's wordpress.

I exposed the wp-cli interface in wp-now so that I can author a plugin that creates a static site from wordpress using ob_start. For instance:

```
 class WDS_CLI {

	public function build() {
		if (!file_exists('dist')) {
			mkdir('dist', 0777, true);
		}
		ob_start();
		// Include the template file
		include 'index.php';

		// End buffering and return its contents
		$output = ob_get_clean();
		$myfile = fopen("dist/index.html", "w") or die("Unable to open file!");
		fwrite($myfile, $output);
		fclose($myfile);		   
	}
}
```

## How?

I added a mode that looks for a wp-content folder without a wp-include folder. From there it was fairly short work to get the mounts correct to use the folder. 

## Testing Instructions

1. Make a site in wordpress-playground and download a .zip.
2. In a checked out copy of playground tools do
  npm run build
3. mkdir test && cd test
3. `unzip ~/Downloads/wordpress-playground\ \(8\).zip` or something
4. run `node ../dist/packages/wp-now/main.js start`

Thanks so much for this package and your advice on using it with my students. My plan is to use wp-playground and the .zip download for my student's first wordpress class on Monday night.


rhildred:
This now works with my wp-bundler wp-cli command. https://github.com/rhildred/wp-bundler. The idea is that I can add the wp-bundler plugin in wordpress playground and it will build on cloudflare pages or github pages with `wp-now  wp wp-bundler build`

packages/wp-now/src/wp-now.ts
```diff
@@ -175,6 +179,46 @@ async function runWpContentMode(
...
+		getWordpressVersionsPath(),
+		wordPressVersion
+	);
+	php.mount(wordPressPath, documentRoot);
+	await initWordPress(php, wordPressVersion, documentRoot, absoluteUrl);
+	fs.ensureDirSync(wpContentPath);
+
+	php.mount(`${projectPath}/wp-content`, `${documentRoot}/wp-content`);
+
+	if (!fs.existsSync(`${projectPath}/wp-content/db.php`)) {
```
adamziel, line 202: I see why you put that in here. Normally `mountSqlitePlugin()` takes care of that, but it also overrides the database file – and in your case the database file is already there. I'd love `mountSqlitePlugin()` to respect any pre-existing databases so that `runWpContentMode()` could work with both a "regular" WordPress and a Playground export. What do you think @sejas @danielbachhuber ?

adamziel:
This is a great use-case @rhildred, thank you for proposing this PR! It marks an important step in making Playground portable between the browser and a local dev environment ❤️  I left an inline note to see if we could make one of the existing modes support that use-case instead of creating a new mode.

PR #148, opened 2024-01-19
Add "playground" keyword to Playground Block.

ndiego:
This is a minor tweak, but I bet that many people will search for "Playground" or type `/playground` when trying to insert the Playground Block. The addition of the `keyword` property in `block.json` makes this possible. 

Currently, you have to search for "WordPress..." for the block to show up.


adamziel (approved):
Lovely, thank you so much!

PR #150, opened 2024-01-26
wp-now: Recognize when running in WebContainers environment and use the correct URL

adamziel:
Description TBD

CleanShot 2024-01-26 at 16 20 59@2x

### DIFF
--- a/packages/wp-now/src/config.ts
+++ b/packages/wp-now/src/config.ts
@@ -77,7 +77,13 @@ async function getAbsoluteURL() {
 		return getCodeSpaceURL(port);
 	}
 	if (isWebContainer()) {
-		return HostURL.parse('http://localhost:' + port).toString();
+		return (
+			HostURL.parse('http://localhost:' + port)
+				.toString()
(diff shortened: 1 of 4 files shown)


packages/wp-now/src/start-server.ts
```diff
@@ -103,6 +103,12 @@ export async function startServer(
 				),
 				body: body as string,
 			};
+
+			data.headers['host'] = options.absoluteUrl
```
adamziel: This should be handled by webcontainers, not wp-now, and removed from here.

packages/wp-now/src/config.ts
```diff
@@ -75,6 +76,9 @@ async function getAbsoluteURL() {
 	if (isGitHubCodespace) {
 		return getCodeSpaceURL(port);
 	}
+	if (isWebContainer()) {
+		return HostURL.parse('http://localhost:' + port).toString();
```
adamziel: This returns a `http://` URL when it should return a `https://` URL

Issue #96, opened 2023-08-09
wp-now: out of memory error when running for many hours


stokesman:
I often have wp-now running for days and after a while—I think usually > 48 hours—I get something like the following:

`Fatal error: Out of memory (allocated 2097152) (tried to allocate 65536 bytes) in /var/www/html/wp-includes/rest-api/endpoints/class-wp-rest-controller.php on line 388`

After that I can't recover the instance. I'm not sure what info I can provide to help illuminate the cause. I tracked down the debug.log in `~/.wp-now/…` but not sure the output before the out of memory error is related.
<details><summary>debug.log portion</summary>

```log
[08-Aug-2023 20:29:16 UTC] PHP Fatal error:  Uncaught TypeError: property_exists(): Argument #1 ($object_or_class) must be of type object|string, array given in /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php:2427
Stack trace:
#0 /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2427): property_exists(Array, 'Name')
#1 [internal function]: WP_SQLite_Translator->{closure}(Array, 0)
#2 /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2430): array_filter(Array, Object(Closure), 1)
#3 /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(2384): WP_SQLite_Translator->strip_sqlite_system_tables(Array)
#4 /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(1381): WP_SQLite_Translator->get_information_schema_query('SELECT TABLE_NA...')
#5 /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(736): WP_SQLite_Translator->execute_select()
#6 /var/www/html/wp-content/plugins/sqlite-database-integration/wp-includes/sqlite/class-wp-sqlite-translator.php(594): WP_SQLite_Translator->execute_mysql_query('SELECT TABLE_NA...')
[...]

PR #152, opened 2024-02-05
WP-NOW: Use rotatePHPRuntime to avoid memory error

sejas:
- Related to https://github.com/WordPress/playground-tools/issues/96

## What?

It uses the new function `rotatePHPRuntime` to mitigate the memory error.

## How?

Update the dependencies and use `rotatePHPRuntime` for the main instance.

## Testing Instructions

* Run:
   * `nvm use && npm install`
   * `npx nx build wp-now`
   * `node dist/packages/wp-now/cli.js start`
   * Observe wp-now has started correctly


packages/wp-now/src/wp-now.ts
```diff
@@ -42,20 +46,23 @@ export default async function startWPNow(
...
-			absoluteUrl: options.absoluteUrl,
-		},
+	const nodePHPOptions: PHPRequestHandlerConfiguration = {
+		documentRoot,
+		absoluteUrl: options.absoluteUrl,
 	};
 
 	const phpInstances = [];
+	const loadRuntime = async () =>
+		await NodePHP.loadRuntime(options.phpVersion);
```
adamziel, line 53: > Property 'loadRuntime' does not exist on type 'typeof NodePHP'.

This GitHub message looks suspicious, otherwise the PR seems to look good.

adamziel:
@brandonpayton this could be interesting for you, and it would also make `wp-now` far more usable.

brandonpayton:
> @brandonpayton this could be interesting for you, and it would also make `wp-now` far more usable.

@adamziel, that sounds good! I can take a look.

packages/wp-now/src/wp-now.ts
sejas: I think the GitHub message was because we weren't using the latest version of Playground dependencies. Now it's gone.

sejas:
@adamziel , @bgrgicak , Now that we updated and adapted the Playground dependencies to the latest version, I could finish this PR. Would you mind reviewing it?

cc: @brandonpayton

packages/wp-now/src/wp-now.ts
```diff
@@ -42,20 +43,23 @@ export default async function startWPNow(
...
-		phpInstances.push(
-			await NodePHP.load(options.phpVersion, nodePHPOptions)
-		);
+		phpInstances.push(new NodePHP(await loadRuntime(), nodePHPOptions));
 	}
 	const php = phpInstances[0];
+	await rotatePHPRuntime({
+		php,
+		recreateRuntime: loadRuntime,
+		maxRequests: 400,
```
adamziel: Maybe let's try `100`? Also, it would be nice to detect out of memory errors and auto-rotate – not that it's a blocker here.

```suggestion
		maxRequests: 100,
```

adamziel (approved):
LGTM!

sejas:
@adamziel , I reverted this PR. As it seems it was returning a `404 File not found` after the first rotation. I'll debug it more and create another PR.

PR #154, opened 2024-02-06
Add WordPress Playground plugin

bgrgicak:
## What?

This adds the `playground` plugin for plugin-preview functionality and renames the plugin to WordPress Playground.

Requires:
https://github.com/WordPress/wordpress-playground/pull/745 .
https://github.com/WordPress/blueprints/issues/49

## Why?

Users want to be able to preview plugins on their own sites before actually committing to installing them live.

## Testing Instructions

- Checkout the project and copy the plugin to your local WordPress install:
```bash
git checkout sm-collector-plugin
cp packages/collector/ [PATH_TO_YOUR_WORDPRESS]/wp-content/plugins/
```
- Open your local WordPress install and activate the WordPress Playground plugin
- Start Playground by clicking on Tools > Sandbox Site in wp-admin
- After the site loads it should have the same content as your site (files and database)

### Testing Previews
- Install this plugin on a site
- Navigate to `plugins` > `add new`
- Search for "Akismet"
- You should now see a `Preview Now` button next to the `Install Now` button.
- Click `Preview now`, and you should see Playground boot your site with the plugin activated.


packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		foreach ($records as $record) {
+			array_push(
+				$sql_dump,
+				sprintf(
+					'INSERT INTO %1$s (%2$s) VALUES (%3$s);',
+					$wpdb->quote_identifier($table),
+					escape_array(
+						array_keys($record)
```
adamziel: Column names are identifiers, but they're escaped as values here. Should `$wpdb->quote_identifier` be used here, too?

packages/playground/src/playground-db.php
```diff
@@ -0,0 +1,116 @@
...
+	);
+	if ($wpdb->last_error) {
+		error_log($wpdb->last_error);
+		return '';
+	}
+	if (!isset($schema['Create Table'])) {
+		return '';
+	}
+	// A query needs to be on a single line
+	return preg_replace("/\r?\n/", "", $schema['Create Table']);
```
adamziel, line 111: nice!

packages/playground/assets/js/playground.js
```diff
@@ -0,0 +1,55 @@
...
+			{
+				step: 'unzip',
+				zipFile: {
+					resource: 'url',
+					url: playground.zipUrl,
+				},
+				extractToPath: '/wordpress',
+			},
+			{
+				step: 'runSql',
```
adamziel, line 23: Lovely! I'd like Playground to do that automatically eventually, but today of course let's stick to an explicit step.

packages/playground/src/playground-db.php
bgrgicak: I used `$wpdb->prepare` here, but `$wpdb->quote_identifier` would be more appropriate here. 

Done in 1eb00ee177dc2371406681802d9c6802ce093a31

bgrgicak: Now when I think of it, we don't need column names or am I missing something?

This example works.

```
		array_push(
				$sql_dump,
				sprintf(
					'INSERT INTO %1$s VALUES (%2$s);',
					$wpdb->quote_identifier($table),
					escape_array(array_values($record))
				)
			);
```

bgrgicak: I removed column names in dd36deba4f2f20766e22fe023417e9e2ce46885e.

bgrgicak:
@adamziel I was able to remove the need for a proxy by fetching blueprints from the backend. [1faf0c5](https://github.com/WordPress/playground-tools/pull/154/commits/1faf0c599167f26f2f1e3f026bc6cd2c21a8cd34)

packages/playground/playground.php
```diff
@@ -0,0 +1,215 @@
...
+ * @param array $plugin The plugin data.
+ * @return array The modified array of plugin action links.
+ */
+function plugin_install_action_links($action_links, $plugin)
+{
+	if (empty($plugin['blueprints'])) {
+		return $action_links;
+	}
+
+	$blueprint = $plugin['blueprints'][0];
```
adamziel: Oooh this is quite tricky!

We can display the Preview button on every plugin, whether it has a Blueprint or not. When you install a plugin, you don't get any initial setup other than what the plugin does on activation. It makes sense to reproduce that experience here.

Then, what should we do when the plugin ships a Blueprint? My gut says _ignore the Blueprint_.

Why?

These Blueprints are written with a vanilla WordPress site in mind. They'll typically create dummy content, change site options, and so on to reach a minimal preview environment. In some cases, like bbpress, that makes a lot of sense even in the context of your site as it doesn't have any forums in place. In other cases, that would only add confusing bloat like lorem ipsum pages when your site already has pages.

Then, applying a Blueprint means the preview experience will be very different from the Installing experience. At best, you won't get the setup wizard. At worst, you'll think "what the... is this plugin going to create 50 dummy posts when I install it? I better go for another one".

So I suggest let's ignore these Blueprints and open a new issue to discuss that with @tellywort, @StevenDufresne, and also some plugin authors.

adamziel:
The plugin worked beautifully with a real WordPress using a local `wp-env`, yay! It failed to transfer the database when I used wp-now, there must be some limitation in the SQLite integration plugin. It's not a blocker, I'm just making a note for posterity. Other than the note I left inline in the diff, everything looks great. Thank you so much @bgrgicak!

adamziel:
I inspected a part of the SQL export that failed in wp-now, it seems like there's a bunch of newlines involved:

CleanShot 2024-02-07 at 13 02 52@2x

adamziel:
This fixed the whitespace but not the import:

```diff
function escape_array($array)
{
	global $wpdb;
	$escaped = array();
	foreach ($array as $value) {
		if (is_numeric($value)) {
			$escaped[] = $wpdb->prepare('%d', $value);
		} else {
-			$escaped[] = $wpdb->prepare('%s', $value);
+			$escaped[] = $wpdb->prepare('%s', str_replace("\n", "\\n", $value));
		}
	}
	return implode(',', $escaped);
}
```

packages/playground/playground.php
bgrgicak: I removed the need for blueprints in [3bdd04a](https://github.com/WordPress/playground-tools/pull/154/commits/3bdd04a1c6584cd4ef442b9a4ce88e8be0c25b87)

bgrgicak:
@adamziel let's address the wp-now issue in a separate PR.

adamziel:
Yup, it's not a blocker at all here, I was just exploring. I isolated the problem and started a new issue here: https://github.com/WordPress/sqlite-database-integration/issues/74

adamziel:
This is looking really good. Thank you @bgrgicak!

PR #163, opened 2024-02-20
Format existing code files and exclude playground/vendor

sejas:
## What?

On https://github.com/WordPress/playground-tools/pull/154 we added the playground plugin with multiple vendor files that `npx nx format` was formatting after each push.

Captura de pantalla 2024-02-20 a las 15 09 46

This PR format existing code files and excludes those in vendor.

* Format code files.
* Excludes packages/vendor/* files to be formatted
* Excludes `.nx/cahce` from git
* Moves `package/playground/phpcs.xml` to the root directory

## Why?

It's better to keep consistency with formatting tools. 

## How?

I executed `npx nx format`. 

## Testing Instructions

- Tests should pass


danielbachhuber (commented):
I was running into this when working on https://github.com/WordPress/playground-tools/pull/161 this morning.

Is there a reason we don't run a linter in CI?

bgrgicak (approved):
Thanks @sejas!

bgrgicak:
@sejas should we move the packages/playground/phpcs.xml file to the root folder?

sejas:
> @sejas should we move the packages/playground/phpcs.xml file to the root folder?

Yes, I think it makes sense to have a unique phpcs.xml at the root directory. I moved it here https://github.com/WordPress/playground-tools/pull/163/commits/7073a53e6f55c8fce5f8590c3a0d480668d62510
Although, I tried to make it work and  phpcs is returning. `ERROR: No sniffs were registered`.
When I used the whole WordPress phpcs.xml it worked well. We should check if we need to modify the phpcs configuration or the issue only happens on my end.

> Is there a reason we don't run a linter in CI?

The project is not 100% well configured and the linter doesn't throw any error, but prettier format the files.
You can try it by breaking the format in any .js file and then running `npm run lint`, you will see 0 errors, but `npx nx format` will format the files.
I'll need more time to fix the linter to make it work properly and then, yes! it would be great to add it to CI GH Actions.
Another possible workaround would be adding husky pre-push hook to format the files.

I also added `.nx/cahce` to .gitignore https://github.com/WordPress/playground-tools/pull/163/commits/081707114ef31a0bcf2ac0f438b351324ff7c2e3. These files weren't created before.

PR #165, opened 2024-02-26
Add Playground footer

### LINKED ISSUE #145: Add "Built with WordPress Playground" link to the WordPress Playground Block
Something like this (bottom right corner):

CleanShot 2024-01-16 at 15 39 30@2x

bgrgicak:
Fixes #145

## What?

This PR adds a footer to the Playground block.

Screenshot 2024-02-26 at 20 54 48

## Why?

The footer should promote WordPress Playground and hopefully increase adoption. 

## How?

By adding a footer element to the block and adding a link to WordPress Playground.

## Testing Instructions

1. Check out the branch.
2. Start dev site `nx dev wordpress-playground-block`
3. On the dev site create a new post and add the WordPress Playground block
4. Confirm that the footer is loaded in the admin menu
5. Save the post and open it
6. Confirm that the footer is loaded on the post page

### DIFF
--- a/packages/interactive-code-block/project.json
+++ b/packages/interactive-code-block/project.json
@@ -14,7 +14,6 @@
 				"cwd": "dist/packages",
 				"commands": [
 					"cp ../../packages/interactive-code-block/README.md ./interactive-code-block/",
-					"rm interactive-code-block/assets/php_5*",
 					"rm interactive-code-block/assets/php_7_0*",
 					"rm interactive-code-block/assets/php_7_1*",
 					"rm interactive-code-block/assets/php_7_2*",
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -9,7 +9,6 @@ import { json } from '@codemirror/lang-json';
 import { php } from '@codemirror/lang-php';
 import {
 	startPlaygroundWeb,
-	login,
 	PlaygroundClient,
 	phpVar,
 } from '@wp-playground/client';
@@ -20,6 +19,7 @@ import {
 	plus,
 	download,
 	cancelCircleFilled,
+	wordpress,
 	edit,
 } from '@wordpress/icons';
 import { FileNameModal } from '../file-name-modal';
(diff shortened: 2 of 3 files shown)


bgrgicak:
@adamziel I'm leaving this as draft until I can get design feedback. Or we can merge it as is if it looks ok.

I moved it to the left to match the WordPress.org footer style.

adamziel:
@bgrgicak this is great, thank you! I'd only make it say "Powered by WordPress Playground", align the text to right, and point the link to `https://w.org/playground`

bgrgicak:
Done!

Screenshot 2024-02-28 at 08 14 24

jarekmorawski:
Hiya! Thanks for the ping. I don't have the full context, but I wonder if the footer shouldn't be detached from the block's content. Perhaps we could display it like a caption in the Image block and make sure it cannot be removed or edited? If we keep it in the site area, it might get in people's way, obscure content, and lead to some frustration.

Putting it outside the block will also make it stand out more. Here's a quick sketch.

adamziel:
@jarekmorawski Thanks! The way I understand @bgrgicak's screenshot is that it is detached, like on your proposal, only it comes with a dark background instead of a transparent one. The transparent background looks interesting, I'm just worried about the contrast between the font and the background. I assume the image's block caption has a way of dealing with this – if there's an easy way to do the same thing here, it could be a nicer solution. If that would require a lot of work, however, let's stick with the dark background. 

About the Playground logo, let's use WordPress logo for now and explore that separately. Once Playground has an official logo, we'll have to update quite a few places.

bgrgicak:
> that it is detached, like on your proposal, only it comes with a dark background instead of a transparent one. 

Yes, technically it's reattached, but visually it looks like a sticky footer from WP-admin.

> The transparent background looks interesting, I'm just worried about the contrast between the font and the background. I assume the image's block caption has a way of dealing with this – if there's an easy way to do the same thing here, it could be a nicer solution.

I don't think that there is an easy way to do this. Having any background color would make it much simpler.

bgrgicak:
I updated the footer to center the content.
@jarekmorawski @adamziel what do you think?

Screenshot 2024-02-28 at 15 26 42

adamziel (approved):
Looks great to me!

jarekmorawski:
Makes sense! If you want to revisit this issue in the future, we'd take a look at the Image block, which syncs with the Global Styles to ensure the caption is clearly visible no matter the background type and color. We'd go that route if we used an SVG for the WP/PG logo.

image

bgrgicak:
Thank you @jarekmorawski for the image block tip.
I managed to reuse the same approach they have with the `--wp--preset--color--contrast` color.

Here are a few examples of how it looks now.

Screenshot 2024-02-29 at 08 02 45
Screenshot 2024-02-29 at 08 02 23
Screenshot 2024-02-29 at 08 01 58

bgrgicak:
@adamziel should we merge this?

What's the release process for a new version?

packages/interactive-code-block/project.json
```diff
@@ -14,7 +14,6 @@
 				"cwd": "dist/packages",
 				"commands": [
 					"cp ../../packages/interactive-code-block/README.md ./interactive-code-block/",
-					"rm interactive-code-block/assets/php_5*",
```
bgrgicak, line 17: @adamziel this line was breaking the build. I assume we don't support PHP 5 anymore, so I removed it.

Let me know if I should move this to a new PR.

adamziel: Good call removing it, it's fine to do it in this PR

adamziel:
I'd make the entire `Powered by WordPress Playground` text a link, otherwise this looks great – thank you @bgrgicak! About the publishing process, it's [described in the README file](https://github.com/WordPress/playground-tools/blob/trunk/packages/wordpress-playground-block/README.md#publishing) that I just updated to make that description clearer. At the moment you'd need SVN repo permissions. We can make it happen, or I could publish it this time, or perhaps it would be more convenient to set up GitHub actions for deployment as @jdevalk described here:

* https://joost.blog/github-actions-wordpress/
* https://github.com/10up/action-wordpress-plugin-deploy

bgrgicak:
Done, @adamziel could you publish it this time? It would be best to automate it in the future instead of giving me permission. 

We could probably reuse the publishing code for the Playground plugin.

adamziel:
Awesome work, thank you @bgrgicak! I'll publish it manually this time. Let's explore the actions once we're publishing frequently enough that handling the SVN workflow requires more time than preparing an action.

PR #167, opened 2024-02-29
Prepare Playground plugin for publishing

bgrgicak:
## What?

Updates the readme to match WordPress.org expectations and adds content export filters to allow skipping of sensitive or unused files. 

## Why?

We want to publish this plugin to WordPress.org.

## How?

By updating the readme and adding filters to the data exporter.

## Testing Instructions

1. Check out the branch.
2. Add the Playground plugin to a WordPress site
3. Follow the Usage instructions in readme.txt


packages/playground/README.txt
```diff
@@ -0,0 +1,57 @@
+=== WordPress Playground ===
+Contributors: antoniosejas, berislavgrgicak, zieladam
```
bgrgicak: @seanmorris what's your WordPress.org username? I would like to include you as a contributor.

packages/playground/src/playground-zip.php
```diff
@@ -32,6 +32,9 @@ function zip_wp_content($zip)
 			continue;
 		}
 
+		if (apply_filters('playground_zip_exclude_file', false, $file)) {
```
adamziel: Or

```suggestion
		$file = apply_filters('playground_exported_file', $file);
		if ( false !== $file ) {
```

This way a plugin could modify the file contents if it wanted to

packages/playground/README.txt
adamziel: Let's start that list with `wordpressdotorg,` similarly to [what the classic-editor plugin does](http://plugins.svn.wordpress.org/classic-editor/trunk/readme.txt).

adamziel: Also let's format this entire file using the wiki syntax with `==` for headers, also similarly to that file referenced above. I'll do the same for block plugin.

packages/playground/README.txt
```diff
@@ -0,0 +1,57 @@
...
+Tested up to: 6.4
+Stable tag: 0.1.0
+Requires PHP: 7.0
+License: GPLv2
+License URI: https://www.gnu.org/licenses/gpl-2.0.html
+Early preview: Enables running a sandbox of your site using WordPress Playground (https://github.com/WordPress/wordpress-playground)
+
+# WordPress Playground
+
+_v0.1.0_
```
adamziel: Let's label this as Changelog and format similarly to the Classic Editor readme http://plugins.svn.wordpress.org/classic-editor/trunk/readme.txt

packages/playground/README.txt
bgrgicak: I updated both readme files

packages/wordpress-playground-block/README.plugindirectory.txt
```diff
@@ -1,5 +1,5 @@
 === WordPress Playground Block ===
-Contributors: dawidurbanski, zieladam
+Contributors: wordpressdotorg, dawidurbanski, zieladam
```
adamziel, line 2: Thanks!

adamziel:
Good work, thank you!

PR #169, opened 2024-03-01
Fix/wp now shared file mounting

### LINKED ISSUE #134: VS Code extension: Cant open anything after updating a plugin
i tried to use it on visual studio code and it doesnt work anymore after updating a plugin

even trying to load a new wordpress folder it shows the same error

Warning: require_once(/var/www/html/wp-content/db.php): failed to open stream: No error information in /var/www/html/wp-includes/load.php on line 671

Fatal error: require_once(): Failed opening required '/var/www/html/wp-content/db.php' (include_path='.:') in /var/www/html/wp-includes/load.php on line 671

bgrgicak:
## What?

This PR disables SQLite plugin updates and ensures that deleted files from the `wp-now` folder are added back on start.

## Why?

By clicking update on the SQLite plugin it would be deleted from the system wp-now folder. This would break wp-now on that machine.

## How?

By disabling updates of the SQLite plugin we will prevent users from breaking their environments.
For users that have broken environments, they will be automatically fixed on start by downloading missing folders.

## Testing Instructions

1. Checkout trunk
2. [Start a local wp-now env](https://github.com/WordPress/playground-tools/tree/trunk/packages/wp-now#contributing)
3. On the site go to the [Plugins screen](http://localhost:8881/wp-admin/plugins.php)
4. Click update on the `SQLite Database Integration` plugin
5. The update should fail and on reload your site will be broken
6. Check out the branch.
7. Rebuild and restart wp-now
8. Confirm that your site is working again
9.  On the site go to the [Plugins screen](http://localhost:8881/wp-admin/plugins.php)
10. Confirm that you can't update the `SQLite Database Integration` plugin


packages/wp-now/src/wp-now.ts
```diff
@@ -361,12 +362,20 @@ function mountMuPlugins(php: NodePHP, vfsDocumentRoot: string) {
...
+		SQLITE_FILENAME
+	);
 	if (php.listFiles(sqlitePluginPath).length === 0) {
-		php.mount(getSqlitePath(), sqlitePluginPath);
+		fs.copySync(getSqlitePath(), sqlitePluginPath);
 		php.mount(
-			path.join(getSqlitePath(), 'db.copy'),
+			path.join(sqlitePluginPath, 'db.copy'),
 			`${vfsDocumentRoot}/wp-content/db.php`
 		);
```
bgrgicak: I think it's best to move this plugin to mu-plugins. This will prevent unwanted updates and will match what hosting platforms do.

bgrgicak: I will keep it as is for now and attempt to fix plugin installing.

adamziel: > I think it's best to move this plugin to mu-plugins. This will prevent unwanted updates and will match what hosting platforms do.

This is what Playground core does, too, to support multisites. Enabling a multisite requires deactivating all plugins, and this is one plugin we cannot deactivate.

bgrgicak:
This PR only prevents users from updating the `SQLite Database Integration` plugin. Because this plugin is shared between sites it shouldn't be updated.

We could consider [copying the plugin to every site ](https://github.com/WordPress/playground-tools/pull/169/commits/56d7ba4094039824d9260586a6f670a3ac874561)instead of mounting it to allow updates in the future.

There is also a [bigger issue with wp-now where any plugin update would break](https://github.com/WordPress/playground-tools/issues/134#issuecomment-1973182771). We will address this issue in a separate PR.

packages/wp-now/src/wp-now.ts
bgrgicak: When I try mounting the plugin to mu-plugins the [file_exists check fails.](https://github.com/WordPress/sqlite-database-integration/blob/8a4efbecf652a2ee3c5facd31ba0542b0d397c13/db.copy#L17) This prevents the DB from loading.

I assume that the root cause of this is the same as the [is dir issue with updates](https://github.com/WordPress/playground-tools/issues/134#issuecomment-1973182771).

```
function mountSqlitePlugin(php: NodePHP, vfsDocumentRoot: string) {
	const sqlitePluginPath = `${vfsDocumentRoot}/wp-content/mu-plugins/${SQLITE_FILENAME}`;
	if (php.listFiles(sqlitePluginPath).length === 0) {
		php.mount(getSqlitePath(), sqlitePluginPath);
		php.writeFile(
			`${vfsDocumentRoot}/wp-content/db.php`,
			php
				.readFileAsText(path.join(getSqlitePath(), 'db.copy'))
				.replace(
					'{SQLITE_IMPLEMENTATION_FOLDER_PATH}',
					sqlitePluginPath
				)
		);
	}
}
```

bgrgicak: We could copy the files, but in that case we would also need to copy the other mu-plugin files.

adamziel: This is how Playground does it:

https://github.com/WordPress/wordpress-playground/blob/8b74852e9701f5083b367f9a294f34200230ce79/packages/playground/wordpress/build/Dockerfile#L36-L42

Perhaps something from that script is missing in the wp-now logic? Ideally, both tools will soon be covered by the PHP Blueprints library.

adamziel:
This would address the problem at hand, but we'll still need to turn the SQLite integration plugin into an mu-plugin to support multisites. What would it take to get the mu-plugin right instead of going through an intermediate step of disabling the updates?

bgrgicak:
@adamziel I moved the SQLite plugin to mu-plugins.

packages/wordpress-playground-block/src/components/playground-preview/index.tsx
```diff
@@ -293,157 +293,174 @@ export default function PlaygroundPreview({
...
-			></iframe>
-		</main>
+				)}
+				<iframe
+					key="playground-iframe"
+					ref={iframeRef}
+					className="playground-iframe"
+				></iframe>
+			</main>
+			<footer className="demo-footer">
```
adamziel: This needs a rebase

packages/wp-now/src/download.ts
```diff
@@ -194,12 +202,32 @@ export async function downloadWordPress(
...
+
+	// Replace {SQLITE_IMPLEMENTATION_FOLDER_PATH} with the actual path
+	fs.writeFileSync(
+		getSqliteDbCopyPath(),
+		fs
+			.readFileSync(getSqliteDbCopyPath())
+			.toString()
+			.replace(
+				"'{SQLITE_IMPLEMENTATION_FOLDER_PATH}'",
+				`realpath( __DIR__ . '/mu-plugins/${SQLITE_FILENAME}-main' )`
```
adamziel: Why use the `-main` suffix?

packages/wp-now/src/get-sqlite-path.ts
```diff
@@ -6,5 +6,12 @@ import { SQLITE_FILENAME } from './constants';
  * The full path to the "SQLite database integration" folder.
  */
 export default function getSqlitePath() {
-	return path.join(getWpNowPath(), `${SQLITE_FILENAME}-main`);
+	return path.join(getWpNowPath(), 'mu-plugins', `${SQLITE_FILENAME}-main`);
```
adamziel: Cound `-main` be a part of `SQLITE_FILENAME`? Right now one has to remember to append it sometimes.

packages/wp-now/src/get-sqlite-path.ts
```diff
@@ -6,5 +6,12 @@ import { SQLITE_FILENAME } from './constants';
...
+}
+
+/**
+ * The full path to the "SQLite database integration" db.copy file.
+ */
+export function getSqliteDbCopyPath() {
```
adamziel, line 15: Nice, no more remembering the path, the function is the source of truth.

packages/wp-now/src/download.ts
```diff
@@ -194,12 +202,32 @@ export async function downloadWordPress(
 }
 
 export async function downloadSqliteIntegrationPlugin() {
-	return downloadFileAndUnzip({
+	// Remove the old SQLite plugin if it exists
+	const oldSqlitePath = path.join(getWpNowPath(), `${SQLITE_FILENAME}-main`);
```
adamziel: Does this download the SQLite integration plugin every time `wp-now` starts? Couldn't it reuse the cached `~/.wp-now` version?

adamziel:
So the main idea here is copying the SQLite integration plugin into `mu-plugins` instead of mounting it as a regular plugin from ~/.wp-now, is that correct? I like the `mu-plugins` part, but I wonder what would it take to preserve the mounting.

packages/wp-now/src/download.ts
bgrgicak: Because this is the plugin name inside mu-plugins.

packages/wordpress-playground-block/src/components/playground-preview/index.tsx
bgrgicak: Fixed

packages/wp-now/src/download.ts
bgrgicak: No, only if it's missing. 

The code you highlighted only ensures that we remove the old folder. 
The folder changed from `.wp-now` to `.wp-now/mu-plugins`.

bgrgicak:
> So the main idea here is copying the SQLite integration plugin into mu-plugins instead of mounting it as a regular plugin from ~/.wp-now, is that correct? I like the mu-plugins part, but I wonder what would it take to preserve the mounting.

The idea is to move the SQLite integration plugin into mu-plugins and mount it from there.
We aren't copying files, they are all mounted from `.wp-now/mu-plugins`

adamziel:
Aha, thank you for explaining! That sounds great 👍

adamziel (approved):
It worked well in my local testing, thank you @bgrgicak! I left a few nitpicks and the testing instructions could use brushing up for posterity, but other than that it's a solid PR.

packages/wp-now/src/get-sqlite-path.ts
bgrgicak: Done

bgrgicak:
@adamziel I addressed the comments and[ fixed an issue with an error notice appearing on the first load of the plugin list. ](https://github.com/WordPress/playground-tools/pull/169/commits/9c4d0f8bc5c6a30d63ad8a897ab29eacb67b7e10#diff-23dc3da15a32706532106007bf3c307e5c622ca35daf548b4931cacb13bff8aeR254-R270)

sejas (approved):
@bgrgicak , this is a great solution, thanks for fixing this issue. I think moving the sqlite plugin to mu-plugins is the right decision.

I followed the testing instructions to test it locally, and I confirm it works as expected. Now the sqlite plugin is a mu-plugin and it cannot be updated, to prevent users from breaking their site.

| **Before** | **After** |
|---|---|
| Captura de pantalla 2024-03-05 a las 15 46 14 | Captura de pantalla 2024-03-05 a las 15 47 11 |

Captura de pantalla 2024-03-05 a las 15 47 47
Captura de pantalla 2024-03-05 a las 15 47 55

PR #183, opened 2024-03-12
wordpress-playground-block: Require Activation Before Loading Playground Iframe

brandonpayton:
## What?

This PR updates the wordpress-playground-block to only load Playground after an Activate button has been clicked.

## Why?

IFrames can create difficulty for those using Screen Readers. See: [#](https://github.com/WordPress/Learn/pull/2264#issuecomment-1971503159), [#](https://github.com/WordPress/Learn/pull/2264#issuecomment-1979277253)

Requiring user action to load Playground will help avoid distracting screen readers with auto-loading iframes. As a bonus, it would also stop the block from automatically consuming network bandwidth to download Playground assets, even if Playground will not be used.

## How?

This button must be clicked before the code editor and Playground load, so the iframe no longer loads immediately.

## Questions

Would it be better to just gate the iframe with an Activate button? Perhaps that doesn't make sense because of the expected interaction with the code editor, but I think it _might_ look better than a single activation button in the middle of a large block like this one.

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -279,7 +279,13 @@ export default function PlaygroundPreview({
 			const lastPath = await playgroundClientRef.current!.getCurrentURL();
 			await playgroundClientRef.current!.goTo(getRefreshPath(lastPath));
 		}
-		doHandleRun();
+
+		if (!isLivePreviewActivated) {
+			// Activate and let the code be run by Playground init
+			setLivePreviewActivated(true);
+		} else {
+			doHandleRun();
+		}
 	}, [reinstallEditedPlugin]);
 
 	const keymapExtension = useMemo(
@@ -300,6 +306,10 @@ export default function PlaygroundPreview({
 		'is-full-width': !codeEditorSideBySide,
 		'is-half-width': codeEditorSideBySide,
 	});
+
+	const iframeCreationWarning = 
+		'This button creates an iframe containing a full WordPress website ' +
+		'which may be a challenge for screen readers.';
 
 	return (
 		<>
@@ -445,6 +455,11 @@ export default function PlaygroundPreview({
 									handleReRunCode();
 								}}
 								className="wordpress-playground-run-button"
+								aria-description={
+									requireLivePreviewActivation
+										? iframeCreationWarning
+										: undefined
+								}
 							>
 								Run
 							</Button>
@@ -457,11 +472,7 @@ export default function PlaygroundPreview({
 							className="wordpress-playground-activate-button"
 							variant="primary"
 							onClick={() => setLivePreviewActivated(true)}
-							aria-description={
-								'This button creates an iframe containing ' +
-								'a full WordPress website which may be ' +
-								'a challenge for screen readers.'
-							}
+							aria-description={iframeCreationWarning}
 						>
 							Activate Live Preview
 						</Button>
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -307,7 +307,7 @@ export default function PlaygroundPreview({
 		'is-half-width': codeEditorSideBySide,
 	});
 
-	const iframeCreationWarning = 
+	const iframeCreationWarning =
 		'This button creates an iframe containing a full WordPress website ' +
 		'which may be a challenge for screen readers.';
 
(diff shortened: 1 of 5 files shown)


brandonpayton:
This is a work in progress. At least the presentation needs improved, if nothing else.

Here's how it currently looks:
image

brandonpayton:
A couple of thoughts on this:

Maybe the entire container should be made clickable to activate with a message like "Click to Activate WordPress Playground" or something like that.

Let's add a block setting like "Require Activation" to control this behavior, defaulting to ON.

brandonpayton:
> Maybe the entire container should be made clickable to activate with a message like "Click to Activate WordPress Playground" or something like that.

I am also exploring whether it makes more sense to just require activation of the Playground iframe itself instead of requiring activation for the whole block.

packages/wordpress-playground-block/src/components/playground-preview/index.tsx
```diff
@@ -296,158 +301,168 @@ export default function PlaygroundPreview({
...
-							>
-								Run
-							</Button>
-						</div>
-					</div>
+				{!isActivated && (
+					<Button
+						className="wordpress-playground-activate-button"
+						onClick={() => setActivated(true)}
+					>Activate WordPress Playground</Button>
```
adamziel:
```suggestion
					>Activate live preview</Button>
```

brandonpayton: Thanks, @adamziel! That describes the intent much better.

Should "live preview" be capitalized like "Activate Live Preview" or do we typically lowercase subsequent words on buttons?

adamziel: I like the capitalization! And let's also add aria attributes to communicate this will create an iframe with a full WordPress site embedded in it, and that it may be a challenge for screen readers.

brandonpayton: :+1: That sounds good.

brandonpayton:
> I am also exploring whether it makes more sense to just require activation of the Playground iframe itself instead of requiring activation for the whole block.

If we do this, then the "Run" button of the code editor doesn't immediately have a loaded iframe to interact with. But we can make the Run button auto-activate the iframe and run when the iframe is loaded.

brandonpayton:
Current rendering when activation is limited to iframe area:
image

The code editor seems like it should be taking the full vertical space down to the bar containing the "Run" button. Will look at fixing this in a separate PR.

And the area with the activation button seems too empty and like its are needs to be more visually distinct from the background. Also, when cursor hovers over the Activation button, the button's background changes to the same color as the container background.

brandonpayton:
Activation button position is currently off in the editor:
image

brandonpayton:
The off-center button was due to a caching issue. The button's background is transparent in the edtor though, and that needs fixed.

I added a block setting "Require live preview activation" and defaulted it to true.
image

Here is the setting when disabled:
image

The language could be improved. So far, I've found it challenging to keep the text descriptive yet brief.

brandonpayton:
I added an ARIA description to the Activate button:
> This button creates an iframe containing a full WordPress website which may be a challenge for screen readers.

It is untested in an actual Screen Reader, so I made a note to do that.

brandonpayton:
What appears to be left on this PR:
- Fix styles
- Consider improved language for new block setting
- Test Activation button with screen reader
- Make code editor Run button trigger activation and run the pending code once the preview is ready

adamziel:
FYI @alexstine. There's still some work left to do here, but I just wanted to share the "activate preview" button for the WordPress Playground block is on its way. Nothing actionable for now :-)

brandonpayton:
> What appears to be left on this PR:
> - Fix styles
> - Consider improved language for new block setting
> - Test Activation button with screen reader
> - Make code editor Run button trigger activation and run the pending code once the preview is ready

These should all be addressed. I believe this PR is ready for review.

adamziel:
This is great, thank you so much @brandonpayton! The only note I have is related to the "Require live preview activation" switch in wp-admin. It works, but changing its value doesn't update the live preview in the editor. I thought it was broken for a second:

CleanShot 2024-03-15 at 13 15 35@2x

Other than that, this is solid. I'll go ahead and merge and let's address the preview problem in a follow up PR. Thank you so much @brandonpayton!

brandonpayton:
> Other than that, this is solid. I'll go ahead and merge and let's address the preview problem in a follow up PR.

Thank you, @adamziel!

> The only note I have is related to the "Require live preview activation" switch in wp-admin. It works, but changing its value doesn't update the live preview in the editor. I thought it was broken for a second:

I wrestled with what to do here and settled on the current, confusing behavior. 😅 I agree that we should fix it and [created an issue](https://github.com/WordPress/playground-tools/issues/192).

PR #188, opened 2024-03-13
VS Code and wp-now: Allow the login() step to fail

adamziel:
The user may customize their login details. Changing the password should not prevent them from starting the server again. This PR adds tolerance for login() failures.

## Other changes

This PR also bumps the VS Code extension version

## Testing instructions

1. Start a wp-now site
1. Change the admin password
1. Restart the site
1. Confirm the server started
1. Confirm it wouldn't start on trunk


adamziel:
I'll go ahead and merge as this issue affects the VS Code extension

PR #190, opened 2024-03-13
WP-NOW: Add GitHub action to release a new version

adamziel:
## What?

Let's make wp-now releases easy with a GitHub workflow.

## Testing instructions

This workflow cannot be tested without merging and running it. Read this PR, confirm it makes sense, and then let's merge and iterate on the workflow.

cc @sejas @bgrgicak

### DIFF
--- a/.github/workflows/release-wp-now.yml
+++ b/.github/workflows/release-wp-now.yml
@@ -2,16 +2,6 @@ name: Release wp-now
 
 on:
     workflow_dispatch:
-        # Require a version bump (patch, minor, major) and an optional dist tag
-        inputs:
-            version_bump:
-                description: 'Version bump (patch, minor, or major)'
-                required: true
-                default: 'patch'
-            dist_tag:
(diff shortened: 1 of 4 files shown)


.github/workflows/release-wp-now.yml
```diff
@@ -0,0 +1,60 @@
...
+        # Require a version bump (patch, minor, major) and an optional dist tag
+        inputs:
+            version_bump:
+                description: 'Version bump (patch, minor, or major)'
+                required: true
+                default: 'patch'
+            dist_tag:
+                description: 'Dist tag (latest, next, etc.)'
+                required: false
+                default: 'latest'
```
sejas: I think we can remove these inputs, since they are not being used.

sejas (approved):
The GitHub action looks good! 👍 
I left a comment about removing the workflow inputs.
I think we can remove them or use them in the last step.

bgrgicak (approved):

adamziel:
Let's give it a try on the next release.

PR #191, opened 2024-03-14
Fix side-by-side code editor to be full height within container

brandonpayton:
## What?

This PR fixes an issue with code editor height. Before this PR, there were some cases where the code editor left some bottom space between it and the Run bar blank and unoccupied.

Before:
before-height-fix

After:
after-height-fix

## Why?

It can be aesthetically displeasing and sometimes confusing to have open, purposeless space within the block.

## How?

This PR changes max-height from
`min(450px, 80vh)`
to
`100%`.

It is possible this is the wrong solution, but I do not currently see a reason we wouldn't want the editor pane to take 100% of the available size.

## Testing Instructions

Check out this branch, follow the [Contributing instructions](https://github.com/WordPress/playground-tools/tree/trunk/packages/wordpress-playground-block#contributing) to build and start a dev version of WordPress, create a post containing the block, and observe the block rendering in the editor and on the front end.

### DIFF
--- a/packages/wordpress-playground-block/src/style.scss
+++ b/packages/wordpress-playground-block/src/style.scss
@@ -141,7 +141,7 @@
 	.cm-editor {
 		width: 100%;
 		height: 100%;
-		max-height: min(450px, 80vh);
+		max-height: 100%;
 	}
 
 	.cm-content {
--- a/packages/wordpress-playground-block/src/style.scss
+++ b/packages/wordpress-playground-block/src/style.scss
@@ -135,13 +135,26 @@
 
 		> div {
 			height: 100%;
+			// Make this a "positioned" element so a nested editor element
+			// may be absolutely positioned relative to this one.
+			position: relative;
 		}
 	}
 
-	.cm-editor {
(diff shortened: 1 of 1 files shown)


brandonpayton:
> This PR changes max-height from
> min(450px, 80vh)
> to
> 100%.
>
> It is possible this is the wrong solution, but I do not currently see a reason we wouldn't want the editor pane to take 100% of the available size.

This is the wrong solution. When testing the editor further, the height just expands indefinitely while I expected it to scroll. Will keep this open and work on a better solution.

brandonpayton:
OK, I see there is also the vertical layout to account for.

brandonpayton:
I've committed a second attempt. This approach absolutely positions the editor element within the editor root in half-width mode which causes it to take exactly the right height, regardless of whether there is overflow. Overflow scrolls and does not affect the height of the editor.

The full-width mode height is left as it was before.

This should be ready for review.

adamziel (approved):
Lovely, thank you @brandonpayton!

PR #197, opened 2024-03-21
playground block: Add initial i18n support

brandonpayton:
## What?

This PR adds initial support for i18n.

I visually scanned PHP, TS, and TSX files and wrapped all the user-facing text I found with calls to `__()`. Because the Playground block is published under the slug `interactive-code-block`, that is the text domain used here.

There are a couple of tasks left:

- [ ] At least some of these need to be converted to `_x()` calls to add context for translators.
- [ ] There is a big block of HTML help text for "Mode" in `edit.tsx` that is not wrapped and needs more careful handling.

## Testing Instructions

1. Check out this branch
2. Run the dev environment with `nx run wordpress-playground-block:dev`
3. Start a new post with localhost:8881/wp-admin/post-new.php
4. Insert a WordPress playground block
5. Enable the Code Editor in the block's settings and poke around in other ways to make sure there are no obvious problems with user facing text.

### DIFF
--- a/packages/wordpress-playground-block/src/block.json
+++ b/packages/wordpress-playground-block/src/block.json
@@ -121,6 +121,12 @@
 	"editorScript": "file:./index.js",
 	"editorStyle": "file:./index.css",
 	"style": ["file:./style-index.css", "wp-components"],
-	"viewScript": ["file:./view.js", "react", "react-dom", "wp-components"],
+	"viewScript": [
+		"file:./view.js",
+		"react",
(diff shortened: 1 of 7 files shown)


brandonpayton:
Note: Any unrelated formatting changes are due to auto-formatting logic that runs on commit.

bgrgicak (approved):
:rocket:

brandonpayton:
This PR had grown stale and is now superceded by #327.

PR #199, opened 2024-03-21
Playground plugin: autologin fix, \n escaping, remove parent admin bar

### LINKED ISSUE #193: incorrect \n parsing. And missing adminbar/not loggedin
Screenshot from 2024-03-17 09-24-51

Viewed together with @adamziel yesterday at Cloudfest.

bgrgicak:
Fixes #193

## What?

This PR fixes a few issues reported in #193.
- Allows users to login if they don't use `admin/password` as login credentials.
- Removes the parent admin bar from the sandbox page.
- Removes `\n` escaping from the DB exporter

### Login

Users with custom usernames and passwords weren't automatically logged into Playground because the credentials didn't match.

### Admin bar

We don't want the parent admin bar to be visible as it would be confusing. 

### \n

Initially, we decided to escape \n as a precaution, but it broke new lines in post content and started displaying them as characters on the page. Escaping isn't necessary and it breaks the export so we can remove it.

### Login

By adding a mu-plugin using blueprints to programmatically login the current user using their id.

### Admin bar

By hiding it with CSS on the Playground admin page.

### \n

By removing the escaping code.

## Testing Instructions

1. Check out the branch.
2. Install the plugin on a local WordPress site or use `wp-env`
3. On the local site update your password to be different from _password_
4. Start a sandbox (wp-admin > Tools > Start Sandbox)
5. Confirm that the sandbox loaded and that you are logged in
6. If you are using the default WP theme open the Sample page by clicking on the link in the header, if not open another page that has multiple paragraphs with line breaks between them.
7. Confirm that there are no `\n` characters

### DIFF
--- a/packages/playground/assets/js/playground.js
+++ b/packages/playground/assets/js/playground.js
@@ -27,7 +27,21 @@
 				},
 			},
 			{
-				step: 'login',
+				step: 'writeFile',
+				path: '/wordpress/wp-content/mu-plugins/0-login.php',
+				data: `<?php
+					add_action( 'setup_theme', function() {
+						if ( is_user_logged_in() ) {
+							return;
+						}
+						$user = get_user_by( 'id', ${playground.userId} );
+						if( $user ) {
+							wp_set_current_user( $user->ID, $user->user_login );
+							wp_set_auth_cookie( $user->ID );
+							do_action( 'wp_login', $user->user_login, $user );
+						}
+					} );
+				`,
 			},
 		],
 	};
@@ -52,10 +66,11 @@
 	});
 
 	await client.isReady();
+	await client.goTo('/');
 
-	if (playground.pluginSlug) {
-		client.goTo('/wp-admin/plugins.php');
-	} else {
-		client.goTo('/wp-admin');
-	}
+	// if (playground.pluginSlug) {
+	// 	await client.goTo('/wp-admin/plugins.php');
+	// } else {
+	// 	await client.goTo('/wp-admin/');
+	// }
 })();
--- a/packages/playground/assets/js/playground.js
+++ b/packages/playground/assets/js/playground.js
@@ -26,23 +26,6 @@
 					path: '/wordpress/schema/_Schema.sql',
 				},
 			},
-			{
-				step: 'writeFile',
-				path: '/wordpress/wp-content/mu-plugins/0-login.php',
-				data: `<?php
-					add_action( 'setup_theme', function() {
-						if ( is_user_logged_in() ) {
-							return;
-						}
-						$user = get_user_by( 'id', ${playground.userId} );
-						if( $user ) {
-							wp_set_current_user( $user->ID, $user->user_login );
-							wp_set_auth_cookie( $user->ID );
-							do_action( 'wp_login', $user->user_login, $user );
-						}
-					} );
-				`,
-			},
 		],
 	};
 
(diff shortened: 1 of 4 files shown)


packages/playground/assets/js/playground.js
```diff
@@ -52,10 +66,11 @@
...
-	if (playground.pluginSlug) {
-		client.goTo('/wp-admin/plugins.php');
-	} else {
-		client.goTo('/wp-admin');
-	}
+	// if (playground.pluginSlug) {
+	// 	await client.goTo('/wp-admin/plugins.php');
+	// } else {
+	// 	await client.goTo('/wp-admin/');
+	// }
```
bgrgicak: @adamziel if I open the homepage or any other non admin page I see the admin bar. If I open `/wp-admin/` I get the login page, but I'm logged in when I click on another page.

I will investigate this tomorrow.

packages/playground/assets/js/playground.js
```diff
@@ -53,9 +50,29 @@
...
+			wp_set_current_user( $user->ID, $user->user_login );
+			wp_set_auth_cookie( $user->ID );
+			do_action( 'wp_login', $user->user_login, $user );
+		}`
+	);
+	await client.request({
+		url: '/playground-login.php',
+	});
+	await client.unlink('/wordpress/playground-login.php');
+
```
bgrgicak, line 72: @adamziel turns out that we need to run the login code using a request to correctly set the cookie.

bgrgicak: This code would also work in the Playground login step if we ever want to support passwordless logins.

packages/playground/assets/js/playground.js
```diff
@@ -66,11 +49,30 @@
...
-	// 	await client.goTo('/wp-admin/plugins.php');
-	// } else {
-	// 	await client.goTo('/wp-admin/');
-	// }
+	// Login as the current user without a password
+	await client.writeFile(
+		'/wordpress/playground-login.php',
+		`<?php
+		require_once( dirname( __FILE__ ) . '/wp-load.php' );
+		if ( is_user_logged_in() ) {
```
adamziel, line 58: Just noting this makes it impossible to log out in the sandbox site. How about basing this on a cookie or a site option instead?

packages/playground/assets/js/playground.js
adamziel: Oh, lovely! Let's track the `login` step idea as a new issue in the https://github.com/WordPress/blueprints-library/ repo – the implementation may differ as we transition to PHP Blueprints.

packages/playground/assets/js/playground.js
bgrgicak: This won't prevent users from logging out. The script is run manually on line 68 and deleted after it runs.
Even if we wouldn't delete the script a user would need to navigate to `/playground-login.php` to run it.

Am I missing something?

packages/playground/assets/js/playground.js
bgrgicak: https://github.com/WordPress/blueprints-library/issues/91

packages/playground/assets/js/playground.js
adamziel: Oh, sorry, you are right!

adamziel:
Let's also throw an error if the plugin is installed on a site using SQLite – there's no support for the required `SHOW CREATE TABLE` there yet and it can be confusing for the user (I just ran into that even though I knew it 😅 )

adamziel (approved):
Works well and solves the issue for me 👍

packages/playground/assets/js/playground.js
```diff
@@ -53,9 +50,29 @@
 
 	await client.isReady();
 
+	// Login as the current user without a password
+	await client.writeFile(
```
brandonpayton, line 54: I really like this approach, @bgrgicak!

PR #201, opened 2024-03-22
WP-NOW: Mount directories after recreateRuntime

sejas:
- Related to https://github.com/WordPress/playground-tools/issues/96
- Follow up of https://github.com/WordPress/playground-tools/pull/152

## What?

A simple attempted on https://github.com/WordPress/playground-tools/pull/200, but we got the `404 File not found` error after the first PHP rotation. I explored more and the documentRoot didn't exist. This PR fixes that by recreating all the necessary directories.

- It uses the new function `rotatePHPRuntime` to mitigate the memory error.
- It mounts the directories following wp-now directives after the php runtime is rotated

## How?

Update the dependencies and use `rotatePHPRuntime` for the main instance.

## Testing Instructions

* Run:
   *  Reduce the maxRequests to 50 or less
   * Run:
   * `nvm use && npm install`
   * `npx nx build wp-now`
   * `node dist/packages/wp-now/cli.js start`
   * Observe wp-now has started correctly, refresh the a few times and observe it loads correctly.

### DIFF
--- a/packages/wp-now/src/wp-now.ts
+++ b/packages/wp-now/src/wp-now.ts
@@ -1,5 +1,5 @@
 import fs from 'fs-extra';
-import { NodePHP, PHPLoaderOptions } from '@php-wasm/node';
+import { NodePHP, PHPRequestHandlerConfiguration } from '@php-wasm/node';
 import path from 'path';
 import { SQLITE_FILENAME } from './constants';
 import {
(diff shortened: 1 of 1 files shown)


sejas:
Keeping the PR as a draft, because I'm getting this error after the first rotation:

```
Error: Could not mount /Users/user/.wp-now/wordpress-versions/latest: Device or resource busy.
    at descriptor.value (/Users/user/playground-tools/node_modules/@php-wasm/node/index.cjs:72642:17)
    at runWpPlaygroundMode (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:765:7)
    at mountDirectories (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:615:13)
    at async loadRuntime (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:651:7)
    at async n (file:///Users/user/playground-tools/node_modules/@php-wasm/universal/index.js:1642:27) {
  [cause]: Error
      at Object.ensureErrnoError (/Users/user/playground-tools/node_modules/@php-wasm/node/index.cjs:29142:33)
      at Object.staticInit (/Users/user/playground-tools/node_modules/@php-wasm/node/index.cjs:29150:10)
      at Object.init4 (/Users/user/playground-tools/node_modules/@php-wasm/node/index.cjs:33265:6)
      at loadPHPRuntime (/Users/user/playground-tools/node_modules/@php-wasm/node/index.cjs:72658:38)
      at _NodePHP.loadRuntime (/Users/user/playground-tools/node_modules/@php-wasm/node/index.cjs:73772:18)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async loadRuntime (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:649:16)
      at async startWPNow (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:656:30)
      at async startServer (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:928:42)
      at async Object.handler (file:///Users/user/playground-tools/dist/packages/wp-now/main.js:1124:25) {
    node: undefined,
    setErrno: [Function (anonymous)],
    errno: 10,
    message: 'FS error'
  }
}
```

adamziel:
This could be solved in Playground core and then all Node.js consumers would benefit.

The runtime swapping logic is implemented in [hotSwapPHPRuntime()](https://github.com/WordPress/wordpress-playground/blob/06f27eb56ac72ba70bdca74f2ee30fb04597e93c/packages/php-wasm/universal/src/lib/base-php.ts#L794).

An easy fix would be to remember all the `mount()` calls in the [NodePHP class](https://github.com/WordPress/wordpress-playground/blob/06f27eb56ac72ba70bdca74f2ee30fb04597e93c/packages/php-wasm/node/src/lib/node-php.ts#L28) and replay them in an overwritten version of `hotSwapPHPRuntime()`. This would only work for NODEFS.

A more complex fix would be to iterate through all the mounts from `FS` associated with the "current" Emscripten PHP instance `FS` and re-apply then to the "next" PHP instance in the original `hotSwapPHPRuntime()` in BasePHP.

----

The `Device or resource busy` message is interesting. I wonder if that directory structure gets recreated in MEMFS and NODEFS refuses to mount it? Or perhaps FS somehow sees that directory is already mounted in another FS instance? Getting to the bottom of this will likely require diving in the php.js Emscripten module.

sejas:
I'll try replaying the mount calls, and probably also the "writeFIles" calls.
I agree that having the solution on Playground core will be a better approach.

adamziel:
> I'll try replaying the mount calls, and probably also the "writeFIles" calls. I agree that having the solution on Playground core will be a better approach.

Writefile is already handled for MEMFS, and NODEFS just uses the disk for storage so the updated data is already available.

PR #157, opened 2024-02-08
Update Playground dependencies to 0.6.7

sejas:
## What?

* Update playground dependencies to 0.6.3
* Adapt wp-now code to use the modified APIs.
  * php.request uses body instead of formData
  * wp-login.php is required to login users through the blueprint step.

## Why?

To remove crypto dependency in node applications

## How?

Updating the main `package.json`

## Testing Instructions

- Run `nvm use && npm install`
- Run `npm run build`
- Run `node dist/packages/wp-now/cli.js start --blueprint=b.json` where b.json is a blueprint with 
- Observe the blueprint executes correctly without any error about `TypeError: crypto.randomUUID is not a function`

```
{
	"$schema": "https://playground.wordpress.net/blueprint-schema.json",
	"landingPage": "/wp-admin/plugins.php",
	"preferredVersions": {
		"php": "8.0",
		"wp": "latest"
	},
	"steps": [
		{
			"step": "login",
			"username": "admin",
			"password": "password"
		},
		{
			"step": "installPlugin",
			"pluginZipFile": {
				"resource": "wordpress.org/plugins",
				"slug": "notification"
			},
			"options": {
				"activate": true
			}
		}
	]
}
```

PR #203, opened 2024-03-27
WP-NOW: fix file uploads using multipart form data

### LINKED ISSUE #202: wp-now: Can't upload media files
## Description

Media files can't be uploaded to sites created with `wp-now`. Uploads fail with the error `Specified file failed upload test.`.

image

## How to reproduce it

> [!NOTE]
> I also reproduce this issue when building `wp-now` locally.

1. Run the command `npx wp-now start`
2. Navigate to WP admin of the site.
3. Navigate to Media page.
4. Upload an image file.
5. Observe the upload fails with the error `Specified file failed upload test.`

sejas:
- Fixes https://github.com/WordPress/playground-tools/issues/202
- Fixes https://meta.trac.wordpress.org/ticket/7535

## What?

This PR fixes the ability to upload Images and any file in general using wp-now.

## Why?

* On https://github.com/WordPress/wordpress-playground/pull/1018/files#diff-71ce1e60e1e8bfb1b2301d9c49d4bd835b28bf65f812e7636c1cae70e4cb5d92 , Playground core v0.6.3 removed the custom `files` property from the request handler.
* wp-now updated the dependencies on https://github.com/WordPress/playground-tools/pull/157 which broke the ability to upload files using the multipart form.

## How?

* In the same PR, Playground core introduced the function `encodeAsMultipart` which was used as a base to fix the issue in this PR.

## Testing Instructions

* Run `nvm use && npm install && npm start`
* Run `npx nx build wp-now && node dist/packages/wp-now/cli.js start`
* Open http://localhost:8881/wp-admin/upload.php
* Click on `Add New Media File`.
* Select any image or valid file like mp3 or mp4.
* Observe the file has been uploaded correctly to the Media Library.

https://github.com/WordPress/playground-tools/assets/779993/f78783f3-8504-456c-bdde-f660837e9d0b

### DIFF
--- a/packages/wp-now/src/encode-as-multipart.ts
+++ b/packages/wp-now/src/encode-as-multipart.ts
@@ -1,5 +1,6 @@
 /**
  * Encodes the Express request with files into multipart/form-data request body.
+ * Adaptation of https://github.com/WordPress/wordpress-playground/blob/45bf16970867cc6f23738224dacf201905adcab6/packages/php-wasm/universal/src/lib/encode-as-multipart.ts
  */
 export async function encodeAsMultipart(req: Express.Request) {
 	const boundary = `----${Math.random().toString(36).slice(2)}`;
--- a/packages/wp-now/src/encode-as-multipart.ts
+++ b/packages/wp-now/src/encode-as-multipart.ts
@@ -1,0 +1,47 @@
+/**
+ * Encodes the Express request with files into multipart/form-data request body.
+ */
+export async function encodeAsMultipart(req: Express.Request) {
+	const boundary = `----${Math.random().toString(36).slice(2)}`;
+	const contentType = `multipart/form-data; boundary=${boundary}`;
+
+	const textEncoder = new TextEncoder();
+	const parts: (string | Uint8Array)[] = [];
+	const data = (req as any).body as Record<string, string>;
+	for (const [name, value] of Object.entries(data)) {
+		parts.push(`--${boundary}\r\n`);
+		parts.push(`Content-Disposition: form-data; name="${name}"`);
+		parts.push(`\r\n`);
+		parts.push(`\r\n`);
+		parts.push(value);
+		parts.push(`\r\n`);
+	}
+	const files = req.files;
+	for (const [name, value] of Object.entries(files)) {
+		if (!Array.isArray(value)) {
+			parts.push(`--${boundary}\r\n`);
+			parts.push(`Content-Disposition: form-data; name="${name}"`);
+			parts.push(`; filename="${value.name}"`);
+			parts.push(`\r\n`);
+			parts.push(`Content-Type: application/octet-stream`);
+			parts.push(`\r\n`);
+			parts.push(`\r\n`);
+			parts.push(value.data);
+			parts.push(`\r\n`);
+		}
+	}
+	parts.push(`--${boundary}--\r\n`);
+
+	const length = parts.reduce((acc, part) => acc + part.length, 0);
+	const bytes = new Uint8Array(length);
+	let offset = 0;
+	for (const part of parts) {
+		bytes.set(
+			typeof part === 'string' ? textEncoder.encode(part) : part,
+			offset
+		);
+		offset += part.length;
+	}
+
+	return { bytes, contentType };
+}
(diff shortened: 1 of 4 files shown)


bgrgicak (approved):

packages/wp-now/src/encode-as-multipart.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Encodes the Express request with files into multipart/form-data request body.
+ * Adaptation of https://github.com/WordPress/wordpress-playground/blob/45bf16970867cc6f23738224dacf201905adcab6/packages/php-wasm/universal/src/lib/encode-as-multipart.ts
+ */
+export async function encodeAsMultipart(req: Express.Request) {
```
adamziel, line 5: Could you just passthrough the raw request bytes to PHP without re-encoding them as multipart?

flexseth:
Niiiiiiiiice 

image

packages/wp-now/src/encode-as-multipart.ts
fluiddot: Good question. I presume we could pass the data using an `object` to let Playground encode it as multipart ([reference](https://github.com/WordPress/wordpress-playground/blob/0df453c11565c05c0c3c9e464459bf9c74677430/packages/php-wasm/universal/src/lib/encode-as-multipart.ts#L7-L47)). The only tricky part would convert the data from `files` property of the request to `File` class so it can be processed properly.

adamziel: @fluiddot the browser has already encoded the request – it should be possible to read the raw request body bytes and pass them, unchanged, to Playground. There's no need to create `File` instances or multipart-encode anything in the JavaScript code.

dcalhoun: I spent time reviewing this logic while [addressing](https://github.com/Automattic/studio/pull/115) a Studio app error, as similar logic is [applied there](https://github.com/Automattic/studio/commit/f9eed4c1bcc57eabfeff09303d26b230c190351b). 

From what I gather, by the time we reach the point to pass the request values onto PHP, the [`express-fileupload`](https://www.npmjs.com/package/express-fileupload) package has relocated any file data from `req.body` to `req.files`.

The file uploading logic was [added](https://github.com/WordPress/wordpress-playground/commit/69aa6c87343954a2f4e341718bfac1abb2ca15b9#diff-5ba071da369e7b99379e26a407462acb97304f8741c748de5deb696c1ecc0472R36) in https://github.com/WordPress/wordpress-playground/pull/188. My perception is that using a package like `express-fileupload` or [`multer`](https://www.npmjs.com/package/multer) is typical and required, as Express does not provide default support for `multipart/form-data`. @sejas does this align with your understanding as well? 

I attempted to merely pass long both `req.body` and `req.files` to PHP as the `body` attribute, but that ended with "out of bounds" errors for the `Uint8Array`, as the files are not in a format that Playground's [multipart encoding](https://github.com/WordPress/wordpress-playground/blob/0df453c11565c05c0c3c9e464459bf9c74677430/packages/php-wasm/universal/src/lib/encode-as-multipart.ts#L7-L47) would recognize currently (instance of `File`). 

I imagine we'd either need to retain this `File` creation/mutlipart-encode logic or expand Playground's multipart-encode logic to accept different file data structures.

adamziel: @dcalhoun here's how `@wp-playground/cli` supports uploads without any multipart parsing:

https://github.com/WordPress/wordpress-playground/blob/946dd7225cec1b5bebbe33072b246609d9e4524a/packages/playground/cli/src/server.ts#L28-L41

adamziel: A multipart-related bug that was submitted recently: https://github.com/WordPress/playground-tools/issues/259

fluiddot: I've tested [the approach](https://github.com/WordPress/wordpress-playground/blob/946dd7225cec1b5bebbe33072b246609d9e4524a/packages/playground/cli/src/server.ts#L28-L41) you shared @adamziel and seems to work if I disable the Express `fileUpload` middleware. I presume the middleware is modifying the request in a way that produces the `out of bounds` error that @dcalhoun described [here](https://github.com/WordPress/playground-tools/pull/203#discussion_r1597233159). I'll create a PR to update this.

fluiddot: [Here's the PR](https://github.com/WordPress/playground-tools/pull/261) with the suggestion applied. Let me know if someone could take a look, thanks 🙇 !

cc @sejas @adamziel

PR #212, opened 2024-04-03
WP-NOW: Add wp-cli integration

### LINKED ISSUE #25: wp-now: Make it possible to use WP-CLI
Once https://github.com/WordPress/wordpress-playground/issues/242 is done, we should add some convenient shorthand for WP-CLI.

**Done is:**
* `wp-now cli` executes WP-CLI.
* If WP-CLI hasn't been downloaded yet, `wp-now` downloads it to its magical hidden folder.

sejas:
- Closes: https://github.com/WordPress/playground-tools/issues/25

## What?

Brings `wp-cli` execution to `wp-now` with the current site context.
The new command will be `wp-now wp ...`
I'm open to other command names, like `cli`.

## Why?

wp-cli is a handy tool for all WordPress developers opening a wide range of automatizations. 

## How?

wp-cli was intentionally removed until the Playground was mature enough to support a wide range of operations.
I added a new command that calls `executeWPCli`, which downloads the phar file and mounts the current directory to make easier the execution of local files with `eval-file`.

## Testing Instructions

1. Run `nvm use && npm install && npx nx build wp-now`
2. Run `npx nx run wp-now:test`
3. Observe the tests pass
4. Run `node dist/packages/wp-now/cli.js wp eval-file /./foo.php`
5. Run `node dist/packages/wp-now/cli.js wp cli version`
6. Run other wp-cli commands
7. Confirm wp-cli runs as expected

### DIFF
--- a/packages/wp-now/src/run-cli.ts
+++ b/packages/wp-now/src/run-cli.ts
@@ -152,24 +152,24 @@ export async function runCli() {
 					process.exit(error.status || -1);
 				}
 			}
-	)
-	.command(
-		'wp',
-		'Run the wp command passing the arguments for wp cli',
-		(yargs) => {
-			return yargs.strict(false);
-		},
-		async () => {
-			// 0: node, 1: wp-now, 2: wp, 3: [wp-cli options...]
-			const args = process.argv.slice(3);
-			try {
-				await executeWPCli(args);
-				process.exit(0);
-			} catch (error) {
-				process.exit(error.status || -1);
+		)
+		.command(
+			'wp',
+			'Run the wp command passing the arguments for wp cli',
+			(yargs) => {
+				return yargs.strict(false);
+			},
+			async () => {
+				// 0: node, 1: wp-now, 2: wp, 3: [wp-cli options...]
+				const args = process.argv.slice(3);
+				try {
+					await executeWPCli(args);
+					process.exit(0);
+				} catch (error) {
+					process.exit(error.status || -1);
+				}
 			}
-		}
-	)
+		)
 		.demandCommand(1, 'You must provide a valid command')
 		.help()
 		.alias('h', 'help')
--- a/packages/wp-now/src/run-cli.ts
+++ b/packages/wp-now/src/run-cli.ts
@@ -8,6 +8,7 @@ import { spawn, SpawnOptionsWithoutStdio } from 'child_process';
 import { executePHP } from './execute-php';
 import { output } from './output';
 import { isGitHubCodespace } from './github-codespaces';
+import { executeWPCli } from './execute-wp-cli';
 
 function startSpinner(message: string) {
 	process.stdout.write(`${message}...\n`);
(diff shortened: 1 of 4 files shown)


packages/wp-now/src/run-cli.ts
```diff
@@ -152,6 +153,23 @@ export async function runCli() {
...
 		)
+		.command(
+			'wp',
+			'Run the wp command passing the arguments for wp cli',
+			(yargs) => {
+				return yargs.strict(false);
+			},
+			async () => {
+				// 0: node, 1: wp-now, 2: wp, 3: [wp-cli options...]
+				const args = process.argv.slice(3);
```
kozer, line 164: Suggestion: Can we replace the 3 with something that express better your comment?
Eg:
``` 
const args = {
    node: 0,
    wpNow: 1,
    wp: 2,
    wpCliOptions: 3
} 
```

kozer:
I tested other commands, and on the first one I got an unhandled exception:
```
❯ node dist/packages/wp-now/cli.js wp cli info
OS:     Emscripten 3.1.43 #1 wasm32
Shell:
PHP binary:
PHP version:    8.0.30-dev
php.ini used:
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "ErrnoError: FS error".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

Node.js v18.18.2
```

Not sure if I needed to pass another argument to make it work, however, for sure the unhandled exception should not be happening. Correct?

packages/wp-now/src/run-cli.ts
adamziel: Wouldn't `yargs` provide you a list of everything that comes after the command?

packages/wp-now/src/tests/wp-now.spec.ts
```diff
@@ -845,4 +845,16 @@ describe('wp-cli command', () => {
 		await executeWPCli(['cli', 'version']);
 		expect(output).toMatch(/WP-CLI (\d\.?)+/i);
 	});
+
+	/**
+	 * Test wp-cli eval-file works correctly.
+	 * We will use the context of Playground mode.
+	 */
+	test('wp-cli eval-file works correctly', async () => {
```
adamziel, line 853: TIL about that command, how lovely!

n3f:
`wp cli info` fails for me too, but it seems to be more of a problem with the PHP wasm implementation than `wp-now`. (At least I couldn't find a way to get it to successfully trap the exception.

adamziel:
@n3f what would be a reproduction?

n3f:
> @n3f what would be a reproduction?

I just followed the instructions in the PR then as an additional case ran `node dist/packages/wp-now/cli.js wp cli info` (just like @kozer).  I saw the same error, but I looked to see if there was a way to catch it, but it appeared to be caught and handled at the php boundary -- but I might just not understand how it works correctly.

sejas:
@n3f , thanks for testing it.
Yeah, not sure why `wp cli info` fails in `php-wasm/node` version and it doesn't in the web version https://playground.wordpress.net/demos/wp-cli.html

Maybe it's a matter of using the latest dependencies.

Screenshot 2024-08-01 at 10 37 59

PR #216, opened 2024-04-03
WP-NOW: Remove message about node v20 limitation for blueprints

sejas:
## What?

Remove message that asks for Node v20 when`--blueprint` param is present.

## Why?

Since the polyfills were implemented on https://github.com/WordPress/wordpress-playground/pull/882 , blueprints can be executed on Node v18

## Testing Instructions

* Create a new blueprint file `b.json` with a valid blueprint like:
```
{
    "landingPage": "/wp-admin/",
    "features": {
        "networking": true
    },
    "steps": [
        {
            "step": "login",
            "username": "admin",
            "password": "password"
        }
    ]
}
```
* Run `nvm use && npm install && npx nx run wp-now:build` # nvm use will set the node version 18.18.2
* Run `node dist/packages/wp-now/with-node-version.js start --blueprint=b.json`
* Observe the blueprint runs as expected and there is no the following error `This script is requires node version v20.0.0 or above when --blueprint=<file> is used; found v18.18.2`

### DIFF
--- a/packages/wp-now/public/with-node-version.js
+++ b/packages/wp-now/public/with-node-version.js
@@ -5,12 +5,8 @@ import fs from 'fs';
 
 // Set the minimum required/supported version of node here.
 
-// Check if `--blueprint=` is passed in proccess.argv
-const hasBlueprint = process.argv.some((arg) => arg.startsWith('--blueprint='));
-
 const minimum = {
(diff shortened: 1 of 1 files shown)


adamziel:
Lovely!

Issue #208, opened 2024-03-29
Playground plugin: Show a useful error when the iframe can't be loaded due to CORP mismatch


bgrgicak:
Some hosts might have Content-Security-Policy headers which would prevent the Playground plugin from loading the sandbox.

We should detect this and show an error to users, so they know why Playground doesn't work.

We should also document this in the readme as a known limitation.

adamziel:
This could be fixed, not just documented. The error message would only be needed to provide instructions about adding a `corp=yes` parameter to the query or so. See:

* https://github.com/WordPress/wordpress-playground/issues/952
* https://github.com/swissspidy/media-experiments/pull/384

PR #219, opened 2024-04-05
Playground plugin: export data and catch load errors

### LINKED ISSUE #213: Playground plugin: Unable to start a sandbox
Hi,
I'm trying to use the _Start a sandbox_ option of the plugin, and ran into a few errors:
1. First, there was #208, so I added `https://playground.wordpress.net/` to my CSP: ✅

2. Then, I got this: `Error: Could not download "https://MYDOMAIN/wp-admin/admin.php?page=playground&download=1"`.
   - I changed the path in [this line](https://github.com/WordPress/playground-tools/blob/e1a9083d818c37fbbbcf220231c0282ad601b5ee/packages/playground/playground.php#L85C16-L85C16) from `admin_url('admin.php?page=' . PLAYGROUND_ADMIN_PAGE_SLUG)`, replacing **admin.php** with **tools.php**: ✅

3. Finally, I got a new error: `Error: Could not read "/wordpress/schema/_Schema.sql": There is no such file or directory OR the parent directory does not exist.`

Not sure how to solve this one :-)

bgrgicak:
Fixes #213 #208

## What?

This PR adds support for exporting Playground data and implements an error modal when Playground fails to load.

## Why?

We would like to allow users to export snapshots of their sites that can later be imported into Playground (a feature in Playground isn't implemented).

Some users reported errors like CORS issues and we need to ensure users have a next step to debug the issue. 

## How?

Playground snapshot is now an option in Tools > Export. 

By catching errors while Playground is loading and displaying an alert. 

## Testing Instructions

1. Check out the branch.
2. Start a site with the plugin running for example by using wp-env
```
cd packages/playground
wp-env start
```
3. Go to wp-admin > Tools > Export
4. Select _Playground snapshot_ and export it
5. Ensure that the downloaded zip contains data
6. Add one of the headers to the top of `playground.php` 
```
header("Content-Security-Policy: default-src 'self'");

header("Content-Security-Policy: default-src 'self' style-src 'self' 'unsafe-inline' 'unsafe-eval'; script-src 'self' 'unsafe-inline' 'unsafe-eval';");
```
7. Try starting a sandbox
8. Confirm that you see an error alert

### DIFF
--- a/packages/playground/assets/js/playground.js
+++ b/packages/playground/assets/js/playground.js
@@ -1,78 +1,94 @@
 (async () => {
-	const { startPlaygroundWeb } = await import(
-		playground.playgroundPackageUrl
-	);
-	const blueprint = {
-		features: {
-			networking: true,
-		},
-		preferredVersions: {
-			wp: playground.wpVersion,
-			php: playground.phpVersion,
-		},
-		steps: [
-			{
-				step: 'unzip',
-				zipFile: {
-					resource: 'url',
-					url: playground.zipUrl,
-				},
-				extractToPath: '/wordpress',
-			},
-			{
-				step: 'runSql',
-				sql: {
-					resource: 'vfs',
-					path: '/wordpress/schema/_Schema.sql',
-				},
-			},
-		],
+	const onError = (error) => {
+		alert(`Playground couldn’t start. Please check the browser console for more information. ${error}`);
+		const backButton = document.getElementById('goBack');
+		if (backButton) {
+			backButton.click();
+		}
 	};
 
-	if (playground.pluginSlug) {
-		blueprint.steps.push({
-			step: 'installPlugin',
-			pluginZipFile: {
-				resource: 'wordpress.org/plugins',
-				slug: playground.pluginSlug,
-			},
-			options: {
-				activate: true,
-			},
-		});
+	if (!window.playground) {
+		onError('Playground script data not found.');
+		return;
 	}
+	try {
+		const { startPlaygroundWeb } = await import(
+			playground.playgroundPackageUrl
+		);
+		const blueprint = {
+			features: {
+				networking: true,
+			},
+			preferredVersions: {
+				wp: playground.wpVersion,
+				php: playground.phpVersion,
+			},
+			steps: [
+				{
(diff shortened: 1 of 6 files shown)


packages/playground/assets/js/playground.js
```diff
@@ -1,78 +1,94 @@
...
+		await client.writeFile(
+			'/wordpress/playground-login.php',
+			`<?php
+			require_once( dirname( __FILE__ ) . '/wp-load.php' );
+			if ( is_user_logged_in() ) {
+				return;
+			}
+			$user = get_user_by( 'id', ${playground.userId} );
+			if( $user ) {
+				wp_set_current_user( $user->ID, $user->user_login );
```
adamziel, line 76: This could be moved to the `login` step in core

packages/playground/assets/js/playground.js
```diff
@@ -1,78 +1,94 @@
...
+				php: playground.phpVersion,
 			},
+			steps: [
+				{
+					step: 'unzip',
+					zipFile: {
+						resource: 'url',
+						url: playground.zipUrl,
+					},
+					extractToPath: '/wordpress',
```
adamziel, line 33: It just struck me that it's writing over the WordPress instance that's already in `/wordpress`. Oh well 🤷  Are we compressing the entire WordPress installation with `wp-admin` and `wp-includes`? I wonder if this step would be more appropriate:

```
    {
      "step": "importWordPressFiles",
      "wordPressFilesZip": {
        "resource": "url",
        "url": playground.zipUrl
      },
      "progress": {
        "weight": 20,
        "caption": "Installing WordPress site"
      }
    }
```

adamziel: Or, even better:

```{
			features: {
				networking: true,
			},
			preferredVersions: {
				wp: playground.zipUrl,
				php: playground.phpVersion
			},
```

adamziel (approved):
I left some notes on how to lean more on Blueprints core, but nothing blocking. LGTM :+1:

bgrgicak:
Thanks! Both make sense and we will need to do it as part of enabling imports in Playground.

Issue #224, opened 2024-04-09
wp-now: Resolve npm audit warnings


eliot-akira:
On installing `@wp-now/wp-now`, there's a warning:

```
4 moderate severity vulnerabilities
```

Running `npm audit report` shows:

```
express  <4.19.2
Severity: moderate
Express.js Open Redirect in malformed URLs - https://github.com/advisories/GHSA-rv95-896h-c2vc
No fix available
node_modules/express
  @php-wasm/node  >=0.1.18
  Depends on vulnerable versions of express
  node_modules/@php-wasm/node
  @wp-now/wp-now  *
  Depends on vulnerable versions of @php-wasm/node
  Depends on vulnerable versions of express
  Depends on vulnerable versions of follow-redirects
  node_modules/@wp-now/wp-now

follow-redirects  <=1.15.5
Severity: moderate
Follow Redirects improperly handles URLs in the url.parse() function - https://github.com/advisories/GHSA-jchw-25xp-jwwc
follow-redirects' Proxy-Authorization header kept across hosts - https://github.com/advisories/GHSA-cxjh-pqwp-8mfp
fix available via `npm audit fix`
node_modules/follow-redirects
```

### Solution

While looking into this, I found that `express` is only used for tests in the entire Playground project.

```
./packages/php-wasm/node/src/test/php-networking.spec.ts:2:import express from 'express';
```

And there are unnecessary dependencies, `express-fileupload` and `follow-redirects`, which are only used for `wp-now` in the Playground Tools project.

```
./packages/wp-now/src/start-server.ts:4:import express from 'express';
./packages/wp-now/src/start-server.ts:7:import fileUpload from 'express-fileupload';

./packages/wp-now/src/download.ts:1:import followRedirects from 'follow-redirects';
```

So the solution is:

- `@php-wasm/node`
  - Update `express` to newest `4.19.2`
  - Move it to `devDependencies`
  - Remove unused `express-fileupload` and `follow-redirects`
- `@wp-now/wp-now`
  - Add `express` to `dependencies`
  - Update `follow-redirects` to newest `1.15.6`

I'll create two pull requests to address these.

PR #225, opened 2024-04-09
Add express to dependencies; Update follow-redirects to newest version

eliot-akira:
## What?

This is the second part of solving:

- #224 

The first part is WordPress/wordpress-playground#1218. (php-wasm/node: Update express to newest version, and move it to devDependencies)

## Why?

The first part of the solution is in the Playground project, which updates `express` and moves it to `devDependencies` since it's only used for tests in that project.

As a result, `express` will no longer be installed when `@php-wasm/node` is installed. So `wp-now` needs to declare `express` as a dependency. 

## How?

- Add `express` to `dependencies`
- Update `follow-redirects` to newest `1.15.6`

These together with WordPress/wordpress-playground#1218 will resolve the original issue #224, which should eliminate all npm audit warnings.

## Testing Instructions

Currently `npm run test` fails with an unrelated issue, as can be seen in another PR's CI test run. https://github.com/WordPress/playground-tools/actions/runs/8616282863/job/23614052137?pr=223#step:4:49 ("RuntimeError: memory access out of bounds") That seems have been caused by commit [133029c](https://github.com/WordPress/playground-tools/commit/133029c236cffa5cee97d599354cb33393b883b6).


sejas:
@eliot-akira , thanks for this PR.
- If we merge https://github.com/WordPress/wordpress-playground/pull/1218 , then adding express as a dependency will be required. 👍 .

If you pull the latest changes from trunk, the tests should pass.

eliot-akira:
OK, merged from trunk, ran `npm install`, and now the tests pass locally and in the CI.

I suppose this PR can be merged before the other one (WordPress/wordpress-playground#1218), since that removes the `express` dependency from `@php-wasm/node`.

As an aside, recently I'm integrating `wp-now` in a team project setup, to make onboarding new members simpler by replacing `wp-env` and Docker. So far it's working great, much faster to get started.

adamziel:
> As an aside, recently I'm integrating wp-now in a team project setup, to make onboarding new members simpler by replacing wp-env and Docker. So far it's working great, much faster to get started.

@eliot-akira YAY! I'm so glad to hear that! Personally I've ran into some rough edges with modes and auto-mounting things and would love to make that optional and configurable via CLI options. But other than it did the trick on my end, too. I think you'd like [Studio](https://developer.wordpress.com/studio/) – @sejas could give you early access.

eliot-akira:
[Studio](https://developer.wordpress.com/studio/) looks cool! From the screenshot I imagine a desktop app for managing local WordPress sites powered by Playground. Curious to try it, I signed up for the waitlist.

I'm still solving a few things to completely replace `wp-env` with `wp-now`.

- The "mapping" feature of `.wp-env.json`, a declarative way to mount local folders to the virtual file system. Surely it can be achieved with blueprints, or a script that reads an existing env file and generates dynamic blueprint steps.
- Run PHPUnit tests using `wp-now`. Getting close but not there yet (ran into WASM error, something to do with `Asyncify.handleSleep` - will file an issue when I have a minimal reproducible example).

adamziel:
@eliot-akira for mapping, let's explore that in Blueprints v2:

https://github.com/WordPress/blueprints-library/issues/46

PR #228, opened 2024-04-10
Playground plugin: tweak the top toolbar

ironnysh:
## What?

This PR changes the message displayed on the top toolbar of a Playground Sandbox. The changes include text, markup, and styles.

The following screenshot demonstrates the changes:
- Top: current toolbar
- Middle: new toolbar
- Bottom: new toolbar when hovering over the **Back** “button”
top-toolbar

## Why?

1. The current color palette is inaccessible (insufficient color contrast).
2. The styles aren't consistent with Playground's existing notice (see [`experimental-notice`](https://github.com/WordPress/wordpress-playground/blob/02a528dd04996d7c9c2a1fdae143ee39c60933f0/packages/playground/website/src/components/browser-chrome/style.module.css#L9)).
3. The copy could be clearer.
4. The `<iframe>` element had no title (required for accessibility reasons).

## How?

I updated two files: the CSS and PHP template:
- Copied [the relevant styles](https://github.com/WordPress/wordpress-playground/blob/02a528dd04996d7c9c2a1fdae143ee39c60933f0/packages/playground/website/src/components/browser-chrome/style.module.css#L12-L13) from the existing notice.
- Updated the link color to pass AA and AAA:
contrast-ratio

- Added styles for the **Back** “button” hover state.
- Changed the HTML markup and copy in the template.

## Testing Instructions

1. Install and activate the Playground plugin
2. Go to `/wp-admin/tools.php?page=playground`
3. See the top toolbar.

### DIFF
--- a/packages/playground/assets/css/playground.css
+++ b/packages/playground/assets/css/playground.css
@@ -14,8 +14,8 @@
 .tools_page_playground #wpbody {
 	position: initial;
 }
-.tools_page_playground #wpadminbar {
-	display: none;
+html.wp-toolbar {
+	padding: inherit;
 }
 
 #wp-playground-toolbar {
--- a/packages/playground/assets/css/playground.css
+++ b/packages/playground/assets/css/playground.css
@@ -20,9 +20,8 @@
 
 #wp-playground-toolbar {
 	background-color: #fff7cc;
-	font-family: "Andale mono", Helvetica, Monospace, Arial;
-	font-weight: 600;
-	padding: 0.5em;
+	font-weight: 500;
+	padding: 0.3em 0.5em;
 	display: flex;
 	gap: 1rem;
 	align-items: center;
@@ -35,17 +34,12 @@
 	transform: translateY(-100%);
 }
 
-#wp-playground-toolbar > a {
-	padding: 0.2rem;
-	color: #1a5687;
-	border: 1px solid;
-	border-radius: 5px;
-	text-decoration: none;
+#wp-playground-toolbar > p {
+	margin: 0;
 }
 
-#wp-playground-toolbar > a:hover {
-	background-color: #1a5687;
-	color: #fff7cc;
+#wp-playground-toolbar > a {
+	color: #1a5687;
 }
 
 #wp-playground-main-area {
--- a/packages/playground/assets/css/playground.css
+++ b/packages/playground/assets/css/playground.css
@@ -19,13 +19,12 @@
 }
 
 #wp-playground-toolbar {
-	background-color: #eaaa00;
-	font-weight: bold;
-	text-align: center;
-	font-size: 1rem;
-	padding: 0.75em;
+	background-color: #fff7cc;
+	font-family: "Andale mono", Helvetica, Monospace, Arial;
+	font-weight: 600;
+	padding: 0.5em;
 	display: flex;
-	flex-direction: row;
+	gap: 1rem;
 	align-items: center;
 	justify-content: center;
 	box-shadow: 0 4px 4px rgba(0, 0, 0, 0.25);
@@ -37,8 +36,16 @@
 }
 
 #wp-playground-toolbar > a {
-	text-transform: capitalize;
-	padding: 0 0.5rem;
+	padding: 0.2rem;
+	color: #1a5687;
+	border: 1px solid;
+	border-radius: 5px;
+	text-decoration: none;
+}
+
+#wp-playground-toolbar > a:hover {
+	background-color: #1a5687;
+	color: #fff7cc;
 }
 
 #wp-playground-main-area {
--- a/packages/playground/templates/playground-page.php
+++ b/packages/playground/templates/playground-page.php
@@ -6,25 +6,25 @@ defined('ABSPATH') || exit;
 ?>
 <div id="wp-playground-wrapper">
     <div id="wp-playground-toolbar">
-        <span>
+        <p>
             <?php
             echo esc_attr(
                 sprintf(
                     // translators: %s: Site name.
                     __(
-                        'WordPress Playground preview for %s',
+                        'WordPress Playground preview of %s',
                         'playground'
                     ),
                     get_bloginfo('name')
                 )
             );
             ?>
-        </span>
+        </p>
         <a href="<?php echo esc_url(admin_url()); ?>" id="goBack">
-            <?php esc_attr_e('Go Back', 'playground'); ?>
+            <?php esc_attr_e('Back', 'playground'); ?>
         </a>
     </div>
     <div id="wp-playground-main-area">
-        <iframe credentialless id="wp-playground"></iframe>
+        <iframe credentialless id="wp-playground" title="WordPress Playground Sandbox"></iframe>
     </div>
 </div>
--- a/packages/playground/templates/playground-page.php
+++ b/packages/playground/templates/playground-page.php
@@ -20,11 +20,11 @@ defined('ABSPATH') || exit;
             );
             ?>
         </p>
-        <a href="<?php echo esc_url(admin_url()); ?>" id="goBack">
+        <a href="<?php echo esc_url(admin_url()); ?>" id="goBack" class="button">
             <?php esc_attr_e('Back', 'playground'); ?>
         </a>
     </div>
     <div id="wp-playground-main-area">
-        <iframe credentialless id="wp-playground" title="WordPress Playground Sandbox"></iframe>
+        <iframe credentialless id="wp-playground" title="<?php _e('WordPress Playground Sandbox', 'playground'); ?>"></iframe>
     </div>
 </div>


packages/playground/templates/playground-page.php
```diff
@@ -6,25 +6,25 @@
...
-        </span>
+        </p>
         <a href="<?php echo esc_url(admin_url()); ?>" id="goBack">
-            <?php esc_attr_e('Go Back', 'playground'); ?>
+            <?php esc_attr_e('Back', 'playground'); ?>
         </a>
     </div>
     <div id="wp-playground-main-area">
-        <iframe credentialless id="wp-playground"></iframe>
+        <iframe credentialless id="wp-playground" title="WordPress Playground Sandbox"></iframe>
```
bgrgicak:
```suggestion
        <iframe credentialless id="wp-playground" title="<?php _e('WordPress Playground Sandbox', 'playground'); ?>"></iframe>
```

packages/playground/assets/css/playground.css
```diff
@@ -19,13 +19,12 @@
 }
 
 #wp-playground-toolbar {
-	background-color: #eaaa00;
-	font-weight: bold;
-	text-align: center;
-	font-size: 1rem;
-	padding: 0.75em;
+	background-color: #fff7cc;
+	font-family: "Andale mono", Helvetica, Monospace, Arial;
```
bgrgicak: Why this font? I would prefer if we stick to the same font like WordPress.

bgrgicak: Also, the font creates some alignment issues that don't exist with the default WordPress font.

image

packages/playground/templates/playground-page.php
```diff
@@ -6,25 +6,25 @@
...
+                        'WordPress Playground preview of %s',
                         'playground'
                     ),
                     get_bloginfo('name')
                 )
             );
             ?>
-        </span>
+        </p>
         <a href="<?php echo esc_url(admin_url()); ?>" id="goBack">
```
bgrgicak:
```suggestion
        <a href="<?php echo esc_url(admin_url()); ?>" id="goBack" class="button">
```

bgrgicak: What if we used the default button style here and just set the background to transparent? 
This would make the size more consistent with the rest of the UI.

packages/playground/assets/css/playground.css
```diff
@@ -37,8 +36,16 @@
 }
```
bgrgicak, line 35: The new bar takes more vertical space than the current version. It would be great to not take up a lot of space because it has very limited value to the user. 

image


```suggestion
}

#wp-playground-toolbar > p {
  line-height: 1;
  margin: 0;
}
```

bgrgicak (commented):
Thank you @ironnysh! This bar bothered me since day one and your direction is a big improvement. I left some comments about the consistency with WP and size.

packages/playground/templates/playground-page.php
ironnysh: Good catch, thanks! :-)

packages/playground/assets/css/playground.css
ironnysh: > Why this font? I would prefer if we stick to the same font like WordPress.

It's the same font list used in the existing notice: https://github.com/WordPress/wordpress-playground/blob/02a528dd04996d7c9c2a1fdae143ee39c60933f0/packages/playground/website/src/components/browser-chrome/style.module.css#L13.

My thinking was to match Playground's existing notice. Didn't think about overall consistency with the WordPress Admin, but it's a good point.
My question would be, do you want it to look as part of the Admin or as part of Playground?

> alignment issues

I'll sort those once there's a decision on the font. Okay?

packages/playground/templates/playground-page.php
ironnysh: Great idea!

But the current link color (`#2271b1`) is still not passing AAA.
Chrome suggested `#1a5687` (what I used in the PR).

packages/playground/assets/css/playground.css
ironnysh: `margin: 0` works, but `line-height: 1;` would be problematic on smaller screens when the site name is long:
image

>  It would be great to not take up a lot of space because it has very limited value to the user.

Speaking of consistency with the existing notice, how about adding the same dismiss button?
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false"><path d="M13 11.8l6.1-6.3-1-1-6.1 6.2-6.1-6.2-1 1 6.1 6.3-6.5 6.7 1 1 6.5-6.6 6.5 6.6 1-1z"></path></svg>`

ironnysh:
> Thank you @ironnysh! This bar bothered me since day one and your direction is a big improvement. I left some comments about the consistency with WP and size.

Thanks, @bgrgicak! I replayed above.

I have to admit, I didn't think about keeping it consistent with WP, but that makes sense :-)

I believe it's a UX question: should this UI element look like it's part of WP's admin or the Playground plugin?

bgrgicak:
> I believe it's a UX question: should this UI element look like it's part of WP's admin or the Playground plugin?

This is all the UI we have for now and it's in WP-admin, so it makes sense to reuse that style.

packages/playground/assets/css/playground.css
bgrgicak: > I'll sort those once there's a decision on the font. Okay?

Sure, I think that using the WP font is the best solution for now.

packages/playground/templates/playground-page.php
bgrgicak: > But the current link color (#2271b1) is still not passing AAA.

Probably because of the background. Would you have any alternative suggestions for how to indicate to the user that they are not on the real site anymore? The bar seems like an overall bad solution and maybe a border around the screen or something else would be enough.

packages/playground/assets/css/playground.css
bgrgicak: This is probably a related discussion https://github.com/WordPress/playground-tools/pull/228#discussion_r1565263791

bgrgicak: What if we remove this bar, and enable the Playground header? This would clearly indicate that the user is inside Playground and they would get all the features that come with that bar.

bgrgicak:
@ironnysh feel free to do whatever you think is best. My only strong opinion is that it should use the least possible amount of space.

packages/playground/assets/css/playground.css
ironnysh: > What if we remove this bar, and enable the Playground header?

That sounds like the best option. But... :-)

> they would get all the features that come with that bar.

Would _everything_ be enabled in this “Sandbox” mode? WP and PHP versions, local storage, reset, download site, etc.?

packages/playground/templates/playground-page.php
ironnysh: If enabling the Playground header isn't an option (due to functionality limitations—see [my questions below](https://github.com/WordPress/playground-tools/pull/228#discussion_r1566206071)), how about adding an admin notice? Something like this:
Display Admin Notice | [Preview](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/adamziel/blueprints/trunk/blueprints/admin-notice/blueprint.json) | [Source](https://github.com/adamziel/blueprints/blob/trunk/blueprints/admin-notice/blueprint.json)
-- | -- | --

ironnysh:
Hi @bgrgicak, so, in retrospect, I agree that getting rid of this element in favor of either of the [options above](https://github.com/WordPress/playground-tools/pull/228#discussion_r1566215121) is the best way forward.

Assuming both are technically viable, let's go with the one that's simpler and more sustainable in the long term (less maintenance in future versions).

If it's an admin notice, I can add it to `playground-page.php`.
Let me know if I can help and/or feel free to close this—it was a fun exploration :-)

packages/playground/templates/playground-page.php
bgrgicak: I'm not sure if a notice is the best solution. A topical admin area is already flooded with notices and they are frequently ignored. I think that the Playground header is the best option.

packages/playground/assets/css/playground.css
bgrgicak: What options would need to be removed? If we need control over that we could add it to Playground, or simply hide buttons with CSS.

ironnysh: It depends on what happens if a user tries to change any of the presets defined in `playground.js`. Would it work as expected, and reload the Sandbox with the new settings?

Assuming the standard Playground header is included in the Sandbox, what happens when I click on the **configurator** button, change something while I'm in `/wp-admin/tools.php?page=playground`, and hit **Apply changes**?

image

bgrgicak: I see now. We can't easily load the Playground website with a header. 
So, seems like we are back to the current header.

@ironnysh would you have time to wrap up this PR and we can merge it into the plugin?

ironnysh: Sure, on it now

ironnysh:
I think we're good to go, @bgrgicak :-)

I checked it on Chromium, Firefox, and Safari in different screen sizes and everything looks aligned (and compact as possible). Let me know if something's broken.

packages/playground/assets/css/playground.css
```diff
@@ -36,9 +34,12 @@
...
 
+#wp-playground-toolbar > p {
+	margin: 0;
+}
+
 #wp-playground-toolbar > a {
-	text-transform: capitalize;
-	padding: 0 0.5rem;
+	color: #1a5687;
```
bgrgicak, line 42:
```suggestion
```

Setting a custom color breaks WordPress theme support.

image

bgrgicak (commented):
Thank you! There are just two small issues left. 

There is a breakpoint where the entire content gets shifted down. I'm not sure if this is related to your PR.

image

packages/playground/assets/css/playground.css
ironnysh: > Setting a custom color breaks WordPress theme support.

I know, but the `.button` class' current color (#2271b1) has insufficient contrast. See the screenshot:

image

ironnysh:
> There is a breakpoint where the entire content gets shifted down. I'm not sure if this is related to your PR.

Nope, it's a weird bug :-) The admin bar is loaded twice. However, it's only visible in 601px-782px because of this CSS line:

```css
@media screen and (max-width:782px)
html.wp-toolbar {
padding-top: 46px;
}
```

The non-media-query CSS should be overridden:

```css
html.wp-toolbar {
padding-top: 32px;
box-sizing: border-box;
-ms-overflow-style: scrollbar;
}
```

I think they are somewhere in core, but couldn't find them.

The good news is that we can fix it by replacing [these lines](https://github.com/WordPress/playground-tools/blob/9c42b734cf7a42c62d788c73bb19edbdb652d44c/packages/playground/assets/css/playground.css#L17-L19) in the plugin's stylesheet with this:

```css
html.wp-toolbar {
	padding: inherit;
}
```

I did a quick check, and seems like it doesn't affect the second “real” admin bar, so I'll commit it.

ironnysh:
> Thank you! There are just two small issues left.

-- @bgrgicak, take 3 is ready :-)

packages/playground/assets/css/playground.css
bgrgicak: Interesting. Does this mean that the color contrast in WP is insufficient?

bgrgicak:
Thank you for working on this @ironnysh!

I will publish it next week.

PR #231, opened 2024-04-12
WP-NOW: Add examples using rewrite-wp-config and WP_DEBUG on Readme.

sejas:
## What?

Update the WP_DEBUG example to use the method `rewrite-wp-config` which avoids conflicting with other constants but it modifies the wp-config.php file shared across projects.

## Why?

The current example works, but it displays warning in the Frontend:

## Testing Instructions

* Read the readme https://github.com/WordPress/playground-tools/blob/update/wp-now-readme-rewrite-wp-config/packages/wp-now/README.md#using-blueprints

### DIFF
--- a/packages/wp-now/README.md
+++ b/packages/wp-now/README.md
@@ -128,7 +128,7 @@ Blueprints are JSON files with a list of steps to execute after starting wp-now.
 
 Here is an example of a Blueprint that defines custom URL constant. `wp-now` will automatically detect the blueprint and execute it after starting the server. In consequence, the site will be available at `http://myurl.wpnow`. Make sure myurl.wpnow is added to your hosts file.
 
-To execute this Blueprint, create a file named `blueprint-example.json` and run `wp-now start --blueprint=path/to/blueprint-example.json`. Note that the `virtualize` is set to `true` to avoid modifying the `wp-config.php` file that is shared between all the projects.
+To execute this Blueprint, create a file named `blueprint-example.json` and run `wp-now start --blueprint=path/to/blueprint-example.json`. Note that the `method` is set to `define-before-run` to avoid modifying the `wp-config.php` file that is shared between all the projects. In contrast, the method `rewrite-wp-config` will modify the `wp-config.php` in disk.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


flexseth:
Thanks for the PR @sejas - the `wp-now` [NPM package page](https://www.npmjs.com/package/@wp-now/wp-now#defining-debugging-constants) may also need update?

flexseth (changes requested):
https://www.npmjs.com/package/@wp-now/wp-now#defining-debugging-constants

Also contains documentation on the constant

bgrgicak (approved):

packages/wp-now/README.md
```diff
@@ -179,18 +179,17 @@ Run `wp-now start --blueprint=path/to/blueprint-example.json` where `blueprint-e
...
 	]
 }
 ```
 
-This will enable the debug logs and will create a `debug.log` file in the `~/.wp-now/wp-content/${project}/debug.log` directory.
+Because we didn't define a method for `defineWpConfigConsts`, and the default method is `rewrite-wp-config`, this blueprint will update your `~/.wp-now/wordpress-versions/latest/wp-config.php` enabling the WP_DEBUG constant and will create a `debug.log` file in the `~/.wp-now/wp-content/${project}/debug.log` directory.
+The next time you execute `wp-now start` in any project, the variable `WP_DEBUG` will still be set to true.
 
-If you prefer to set a custom path for the debug log file, you can set `WP_DEBUG_LOG` to be a directory. Remember that the `php-wasm` server runs udner a VFS (virtual file system) where the default documentRoot is always `/var/www/html`.
+If you prefer to set a custom path for the debug log file, you can set `WP_DEBUG_LOG` to be a directory. Remember that the `php-wasm` server runs udner a VFS (virtual file system) where the default documentRoot for wp-now is always `/var/www/html`.
```
wojtekn:
```suggestion
If you prefer to set a custom path for the debug log file, you can set `WP_DEBUG_LOG` to be a directory. Remember that the `php-wasm` server runs under a VFS (virtual file system) where the default documentRoot for wp-now is always `/var/www/html`.
```

wojtekn (commented):
It seems odd that it's even possible to update the vanilla WP in `~/.wp-now/wordpress-versions/latest/` when we start the server. Shouldn't the downloaded WP files be considered read-only?

packages/wp-now/README.md
sejas: Thanks for the suggestion on the typo.

Yes, I agree that it's odd and it would be awesome to keep the wp version intact 🤔 .
We can probably copy the whole wp directory in VFS and do not mount it, to avoid conflicts.

Anyway, this PR only document the current behaviour.

PR #257, opened 2024-05-08
Playground Block: Move live preview toggle to general panel

### LINKED ISSUE #255: Playground Block: Extract the " Require live preview activation" setting to the top level
The "Require live preview activation" is hidden in the code editor tab and thus difficult to spot when not using the code editor. Let's make it a top-level setting:

CleanShot 2024-05-07 at 12 26 18@2x

bgrgicak:
Fixes #255 

## What?

Moves live preview toggle to a new General panel to make easily discoverable. 

## Why?

The live preview toggle is hidden in the Code editor panel and only accessible if the code editor is enabled.

## How?

By creating a new panel and moving the toggle to it.

## Testing Instructions

1. Check out the branch.
2. Start editing a post and add a new Playground block
3. In Block settings confirm that there is a new Greneral tab and that it has a _Require live preview activation_ toggle

image

### DIFF
--- a/packages/wordpress-playground-block/src/edit.tsx
+++ b/packages/wordpress-playground-block/src/edit.tsx
@@ -59,6 +59,23 @@ export default function Edit({
 			/>
 			<InspectorControls>
 				<Panel header="Settings">
+					<PanelBody title="General" initialOpen={true}>
+						<ToggleControl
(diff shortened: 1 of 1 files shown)


brandonpayton (commented):
I have not tested, but this looks good to me.

PR #258, opened 2024-05-08
Playground block: Base64 encode attributes to prevent KSES from breaking the values on save

adamziel:
Some WordPress installations are overly eager with their HTML entity encoding and will save, e.g., `<?php` as `&lt;php`. We cannot easily detect this to decode these HTML entities only when needed, so let's just store the attributes using base64 encoding to prevent WordPress from breaking them.

This PR encodes the most entity-encoding-susceptible attributes to base64 in the block editor. To avoid an expensive encode/decode operation on each key stroke, the base64 encode operation is debounced.

 ## Testing instructions

* Install the trunk version of this block on a site
* Insert it on a page with a code editor, a few files, and a Blueprint
* Switch to this branch
* Confirm the original block still works in the editor and on the frontend
* Update any attribute, save it, refresh the page
* Confirm it still works in the editor and on the frontend
* Insert a new block, insert a few files in the code editor, update the Blueprint attribute, save it
* Confirm it works in the editor and on the frontend

### DIFF
--- a/packages/wordpress-playground-block/src/base64.ts
+++ b/packages/wordpress-playground-block/src/base64.ts
@@ -1,0 +1,95 @@
+/**
+ * Base64 encoding and decoding functions.
+ * We cannot just use `btoa` and `atob` because they do not
+ * support Unicode characters.
+ */
+
+const attributesToBase64 = [
+	'blueprint',
+	'blueprintUrl',
+	'codeEditorErrorLog',
+	'constants',
+	'files',
+];
+
+export function base64EncodeBlockAttributes(
+	blockAttributes: Record<string, any>
+) {
+	const base64Props: Record<string, string> = {};
+	for (const key in blockAttributes) {
+		if (
+			!attributesToBase64.includes(key) ||
+			typeof blockAttributes[key] === 'number' ||
+			typeof blockAttributes[key] === 'boolean' ||
+			typeof blockAttributes[key] === null ||
+			typeof blockAttributes[key] === undefined
+		) {
+			base64Props[key] = blockAttributes[key];
+			continue;
+		}
+		try {
+			base64Props[key] = stringToBase64(
+				JSON.stringify(blockAttributes[key])
+			);
+		} catch (error) {
+			base64Props[key] = blockAttributes[key] as string;
+		}
+	}
+	return base64Props;
+}
+
+export function base64DecodeBlockAttributes(
+	base64Attributes: Record<string, any>
+) {
+	const attributes: Record<string, any> = {};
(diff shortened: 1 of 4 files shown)


packages/wordpress-playground-block/src/edit.tsx
```diff
@@ -17,8 +18,103 @@ import {
...
+			if (ref.current.encodeTimeout) {
+				clearTimeout(ref.current.encodeTimeout);
+			}
+			ref.current.encodeTimeout = setTimeout(() => {
+				props.setAttributes(
+					base64EncodeBlockAttributes(newBase64Attributes)
+				);
+				clearTimeout(ref.current.encodeTimeout);
+				ref.current.encodeTimeout = null;
+			}, 100);
```
dmsnell, line 101: was this a performance issue? we've introduced a concurrency issue here on `save()`, where we might save the post before this runs and then lose our updates.

packages/wordpress-playground-block/src/base64.ts
```diff
@@ -0,0 +1,106 @@
...
+		) {
+			attributes[key] = base64Attributes[key];
+			continue;
+		}
+		if (key in base64Attributes) {
+			try {
+				attributes[key] = JSON.parse(base64ToString(valueToDecode));
+			} catch (error) {
+				// Ignore errors and keep the base64 encoded string.
+				attributes[key] = base64Attributes[key];
```
dmsnell, line 74: this looks like it should preserve previous attributes which aren't base-64-encoded, but perhaps accidentally so.

I think it's best if we do that: return the raw string if it doesn't base-64-decode, but it could be nice indicating that here.

packages/wordpress-playground-block/src/edit.tsx
adamziel: @dmsnell Yeah I saw slowdown in typing when encoding on each keystroke. You are right this could lead to losing updates, though. I used a timeout small enough to hopefully avoid losing changes in most cases, but it will probably happen to someone eventually. Do you have any alternative ideas?

packages/wordpress-playground-block/src/base64.ts
adamziel: Done in https://github.com/WordPress/playground-tools/commit/b189b9fa47e830320710e848f186922299c59edf

packages/wordpress-playground-block/src/edit.tsx
dmsnell: it doesn't resolve the concurrency issue, but we could run the computation in a WebWorker, which should be faster.

this may be fine enough for now.

PR #274, opened 2024-05-22
Playground block: Strengthen CSS for activate button positioning

### LINKED ISSUE #272: Playground block: Activate button going out of viewport
When there's a lot of code, the activate preview button is vertically centered relative to the code editor height, which can make it go off screen:

CleanShot 2024-05-21 at 17 40 51@2x

cc @brandonpayton

brandonpayton:
## What?

This plugin strengthens the margin property of the Playground block's activate button.

## Why?

Today, when running with the p2020 theme on WordPress.com, another rule sets `margin-bottom: 0` on top of our `margin: auto` and causes the button to be positioned at the bottom of the preview pane.
CleanShot 2024-05-21 at 17 40 51@2x

After this change, the button renders in the context of the p2020 theme without issue:
Screenshot 2024-05-21 at 10 39 23 PM

Fixes #272

## How?

This PR makes `margin: auto` `!important`. It's a shame to need to do this, but AFAIK, there is no scenario in which we want a theme to be able to override the position of the Activate button.

## Testing Instructions

Because this is in the context of a theme running on a private site, it cannot be tested here, but I have tested this change privately, which is how the screenshot was taken.

### DIFF
--- a/packages/wordpress-playground-block/src/style.scss
+++ b/packages/wordpress-playground-block/src/style.scss
@@ -184,7 +184,9 @@
 	}
 
 	.wordpress-playground-activate-button {
-		margin: auto;
+		// Use !important to defend against this button being
+		// incorrectly positioned due to opinionated theme styles.
+		margin: auto !important;
 		font-size: 20px;
 	}
 


adamziel:
I wonder if shadow DOM or iframing the block would solve this entire class of issues.

brandonpayton:
I was thinking that too but talked myself out of it because it seemed like we would still want to be affected by some core and theme styles. If that is not the case, using shadow DOM to isolate this block's content from external styles sounds like a great idea.

brandonpayton:
Created an issue for trying style isolation using shadow DOM: #277

PR #281, opened 2024-05-22
Playground block: Prevent the CodeMirror keyboard shortcuts from leaking to the block editor

adamziel:
## What?

Prevents keyboard events from leaking from the code editor to the block editor. It solves the problem of, e.g., undoing block changes when pressing `cmd+z` with the intention of undoing code changes.

## Testing Instructions

1. Add a few paragraph blocks
2. Add the Playground block
3. Enable the code editor
4. Start typing
5. Press cmd+s, the code should be re-run
6. Press cmd+z a few times, it should only undo the changes inside the code editor but not in the block editor

### DIFF
--- a/packages/interactive-code-block/src/lib/components/code-mirror/index.tsx
+++ b/packages/interactive-code-block/src/lib/components/code-mirror/index.tsx
@@ -51,6 +51,11 @@ export type CodeMirrorRef = {
 	getContents: () => string;
 };
 
+function stopPropagation(event: React.KeyboardEvent) {
+	event.preventDefault();
+	event.stopPropagation();
+}
+
 const CodeMirror: MemoExoticComponent<
 	ForwardRefExoticComponent<CodeMirrorProps & RefAttributes<CodeMirrorRef>>
 > = memo(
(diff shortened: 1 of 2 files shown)


packages/interactive-code-block/src/lib/components/code-mirror/index.tsx
```diff
@@ -150,13 +155,40 @@ const CodeMirror: MemoExoticComponent<
...
+				doc.removeEventListener('keyup', keyListener, true);
+				doc.removeEventListener('keypress', keyListener, true);
+			};
+		}, []);
+
+		return (
+			<div
+				ref={codeMirrorRef}
+				className={className}
+				onKeyDown={stopPropagation}
```
brandonpayton: Why are we adding listeners both to this element and the owner document? Wouldn't the listeners added to the owner document capture phase catch everything?

adamziel: I changed this file by mistake, it's a different block :-) Let me revert this file

brandonpayton: Ah, cool :-) This PR a great change, btw.

adamziel: Thanks!

PR #284, opened 2024-05-31
Allow express app to be (optionally) passed to start-server

Shelob9:
## What?

- Add optional `app` arg to `startServer` that is express app.

## Why?

I wanted to see if I could integrate this with an existing express app. I was able to make it work with one small change that I think is a useful change. Would allow using WordPress for part of a node application and or easily plug this package into existing application

This is what I am thinking:
```ts
import {
	getWpNowConfig,
	startServer
} from '@wp-now/wp-now';

const options = await getWpNowConfig({});
const app = express();
app.get('/api', (req, res) => {
	res.json({message:"Hi Roy"})
});
const {  url } = await startServer({...options,app});
console.log({url})
```
## How?

- Makes it optional arg
- Uses that app if needed

### No Changes test
1. Check out the branch. 
2. `npx nx preview wp-now start`
3. Should work as-is, no changes.

### Works this way test
1. Check out the branch. 
2. Link `npm link @wp-now/wp-now` https://docs.npmjs.com/cli/v9/commands/npm-link
3. Make a new project
4. Install this wp-now locally
5. Test it with above example code

### DIFF
--- a/packages/wp-now/src/config.ts
+++ b/packages/wp-now/src/config.ts
@@ -13,6 +13,7 @@ import { isValidWordPressVersion } from './wp-playground-wordpress';
 import getWpNowPath from './get-wp-now-path';
 import { DEFAULT_PHP_VERSION, DEFAULT_WORDPRESS_VERSION } from './constants';
 import { isWebContainer, HostURL } from '@webcontainer/env';
+import express from 'express';
 
 export interface CliOptions {
 	php?: string;
@@ -47,6 +48,7 @@ export interface WPNowOptions {
 	blueprintObject?: Blueprint;
 	reset?: boolean;
 	landingPage?: string;
+	app?: express.Express
 }
 
 export const DEFAULT_OPTIONS: WPNowOptions = {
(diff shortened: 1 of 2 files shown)


bgrgicak:
Hey @Shelob9, thanks for working on this. While this could work there is a much better way to do it with Playground instead of WP-NOW.
[Playground has a CLI server](https://github.com/WordPress/wordpress-playground/tree/trunk/packages/playground/cli) and it would be great if there was an Express middleware that could run Playground CLI requests.
Would you be up for working on this in the Playground repository?

FYI @juanmaguitar

Shelob9:
>  While this could work there is a much better way to do it with Playground instead of WP-NOW.

Why is this better? The reason I chose to do it this way was:
* This allows declaring routes before the `/` route is added by wp-now
* The function I editted is exported.

bgrgicak:
> Why is this better?

WP-now is a tool that uses Playground and is built for local development. 
Playground CLI is the standard way of running Playground servers. By contributing to the Playground CLI everyone gets the benefits.

adamziel:
A middleware could be used in any express server, while hooking into wp-now makes it specific to the way wp-now is booted and configured.

PR #295, opened 2024-06-04
Rename the plugin to Sandbox, update description, add "experimental" to UI buttons

adamziel:
## What?

Calling every plugin "Playground" is confusing, there's Playground block, Playground plugin, Playground integration etc. This PR renames the Playground plugin to "Sandbox" and clarifies that features are experimental.

cc @bgrgicak

### DIFF
--- a/packages/playground/playground.php
+++ b/packages/playground/playground.php
@@ -1,9 +1,8 @@
 <?php
-
 /**
- * Plugin Name: Playground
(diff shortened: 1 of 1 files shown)


flexseth:
"Sandbox" plugin has not been updated in 10 years
https://wordpress.org/plugins/sandbox/

+1 like the idea

PR #299, opened 2024-06-12
Playground block: Stop monopolizing main element

### LINKED ISSUE #292: Playground block: A11Y: Remove duplicate main landmarks

Every time the Playground block is used, it renders a `<main>` landmark. Really, this landmark should only be used once per page. Can this be replaced by another tag? Maybe just a generic `<div>`? You could also replace this with a `<section aria-label="Playground">` but I think if the Playground block has surrounding headings, it's not a big deal.

brandonpayton:
## What?

Fixes #292 - "Playground block: A11Y: Remove duplicate main landmarks"

This PR replaces the Playground block's `<main>` element with a `<section>` element annotated with an `aria-label`.

## Why?

There is only supposed to be one `<main>` element per document. We do not know if a document including a Playground block will be rendered within another `<main>` element, and there can be multiple Playground blocks in a single document.

## Testing Instructions

- Create a new post
- Add a Playground block in the editor and confirm it renders normally
- Publish the post and confirm the Playground block renders normally on the front end
- Use screen reader and confirm that the section label is noted when engaging with the section

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -354,7 +354,10 @@ export default function PlaygroundPreview({
 
 	return (
 		<>
-			<section aria-label="Playground" className={mainContainerClass}>
+			<section
+				aria-label="WordPress Playground"
+				className={mainContainerClass}
+			>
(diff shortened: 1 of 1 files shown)


brandonpayton:

PR #300, opened 2024-06-12
Playground block: Add ARIA labels to icon-only buttons

### LINKED ISSUE #286: Playground block: A11Y: Unlabeled button in code editor

There is an unlabeled button with icon only, not sure what it should be called.

```
<button type="button" class="components-button file-tab file-tab-extra is-secondary"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><path d="M18 11.3l-1-1.1-4 4V3h-1.5v11.3L7 10.2l-1 1.1 6.2 5.8 5.8-5.8zm.5 3.7v3.5h-13V15H4v5h16v-5h-1.5z"></path></svg></button>
```

brandonpayton:
## What?

This PR adds ARIA labels to the code editor's "Add File" and "Download Code as a Zip File" buttons which previously only included visual indications of their function.

Fixes #286

## Why?

Without ARIA labels, people using assistive technology may have no way to know what the buttons do.

## Testing Instructions

- Create a new post
- Add a Playground block in the editor
- Use a screen reader to engage with the "Add File" and "Download Code as Zip File" button and confirm the labels are recognized
- Publish the post and use a screen reader to confirm the "Download Code as Zip File" button has a recognized label

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -399,6 +399,7 @@ export default function PlaygroundPreview({
 							)}
 							{showAddNewFile && (
 								<Button
+									aria-label="Add File"
 									variant="secondary"
 									className="file-tab file-tab-extra"
 									onClick={() =>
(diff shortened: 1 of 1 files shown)


brandonpayton:

PR #323, opened 2024-07-03
Upgrade @php-wasm/node to 0.6.16

zaguiini:
## What?

Bumps `@php-wasm/node` to the last version before breaking changes.

## Why?

The main reason I'm bumping this is to catch up with the error handling included in `0.6.13...0.6.16`, which improves the developer experience by not crashing the entire process and just reporting the error to the user in the output, allowing them to change the code and refresh the page instead of restarting the server.

## Testing Instructions

- Check all tests are passing;
- Check that you can run `wp-now` with `node <repo-root>/dist/packages/wp-now/main.js start` after building it locally;
- Check that throwing an error within the WP installation crashes the app but not the server;
- Check that removing the error and refreshing the page works.


zaguiini:
cc @sejas

sejas (approved):
I tested wp-now introducing an error and I confirm now the server is still running even with php errors. After fixing the intentional php error, I don't need to restart wp-now.



https://github.com/WordPress/playground-tools/assets/779993/b153baed-2250-48d1-9805-92119cfca285

PR #324, opened 2024-07-05
Download exact WordPress version when passing an alias

### LINKED ISSUE #214: wp-now: "latest" WP never updates
I'm starting wp-now like this:

```bash
wp-now start --reset --blueprint=blueprint-wp-now.json
```

I first ran this when the latest WordPress was 6.4, and it correctly downloaded and used 6.4. Since then WP6.5 was rolled out, but wp-now continues to use 6.4:

```
$ wp-now start --reset --blueprint=blueprint-wp-now.json
Starting the server......
directory: /Users/lowey/devel/squelch-tabs-and-accordions-shortcodes
mode: plugin
php: 7.4
wp: latest
…
```

Removing the "latest" directory before running wp-now triggers a re-download of "latest", bringing it up to the correct version:

```
$ rm -rf ~/.wp-now/wordpress-versions/latest
$ wp-now start --reset --blueprint=blueprint-wp-now.json
Starting the server......
directory: /Users/lowey/devel/squelch-tabs-and-accordions-shortcodes
mode: plugin
php: 7.4
wp: latest
Downloading WordPress latest...
…
```

Obviously it's good that wp-now doesn't download WordPress on every run, but it could do with checking whether what is in the "latest" dir matches what the update API says is the latest version.

zaguiini:
- Closes: https://github.com/WordPress/playground-tools/issues/214

## What?

Resolves to an actual WordPress version when starting `wp-now` with `--wp=latest|trunk|nightly`.

It keeps working while offline by caching the version aliases in a file: `~/.wp-now/wp-version-offline-alias.json`

## Why?

In https://github.com/WordPress/playground-tools/issues/214, it was reported that running `--wp=latest` might start the server with an outdated version of WordPress because we're caching the version locally by the alias (which changes between published versions) and not the version number.

## How?

Instead of having `latest`, `trunk` and `nightly` folders, we're resolving these aliases using the `api.wordpress.org` endpoint to actual versions, and caching the version number instead of the alias.

After confirmation this is the approach we want to move forward with, I'll add/update tests.

## Testing Instructions

Build it locally and verify that following commands produce the following outputs:

| Command | CLI options output |
| ----------- | ------- |
| `start` | `wp: 6.5.5 (resolved from alias: latest)` |
| `start --wp=latest` | `wp: 6.5.5 (resolved from alias: latest)` |
| `start --wp=trunk` | `wp: 6.7-alpha-xxxx (resolved from alias: trunk)` |
| `start --wp=nightly` | `wp: 6.7-alpha-xxxx (resolved from alias: nightly)` |
| `start --wp=6.3` | `wp: 6.3` |

### DIFF
--- a/packages/wp-now/src/download.ts
+++ b/packages/wp-now/src/download.ts
@@ -5,18 +5,12 @@ import { IncomingMessage } from 'http';
 import os from 'os';
 import path from 'path';
 import unzipper from 'unzipper';
-import {
-	DEFAULT_WORDPRESS_VERSION,
-	SQLITE_FILENAME,
-	SQLITE_URL,
-	WP_CLI_URL,
-} from './constants';
+import { SQLITE_FILENAME, SQLITE_URL, WP_CLI_URL } from './constants';
 import getSqlitePath, { getSqliteDbCopyPath } from './get-sqlite-path';
 import getWordpressVersionsPath from './get-wordpress-versions-path';
 import getWpCliPath from './get-wp-cli-path';
 import getWpNowPath from './get-wp-now-path';
 import { output } from './output';
-import { isValidWordPressVersion } from './wp-playground-wordpress';
 
 function httpsGet(url: string, callback: Function) {
 	const proxy =
@@ -37,15 +31,11 @@ function httpsGet(url: string, callback: Function) {
 	https.get(url, { agent }, callback);
 }
 
-function getWordPressVersionUrl(version = DEFAULT_WORDPRESS_VERSION) {
-	if (!isValidWordPressVersion(version)) {
-		throw new Error(
-			'Unrecognized WordPress version. Please use "latest", "trunk", "nightly", or numeric versions such as "6.2", "6.0.1", "6.2-beta1", or "6.2-RC1"'
-		);
-	}
-	if (version === 'trunk' || version === 'nightly') {
+function getWordPressVersionUrl(version: string) {
+	if (version === 'nightly') {
 		return 'https://wordpress.org/nightly-builds/wordpress-latest.zip';
 	}
+
 	return `https://wordpress.org/wordpress-${version}.zip`;
 }
 
@@ -178,12 +168,15 @@ async function downloadFileAndUnzip({
 }
 
 export async function downloadWordPress(
-	wordPressVersion = DEFAULT_WORDPRESS_VERSION
+	wordPressVersion: string,
+	{ isDeveloperBuild = false } = {}
 ) {
 	const finalFolder = path.join(getWordpressVersionsPath(), wordPressVersion);
 	const tempFolder = os.tmpdir();
 	const { downloaded, statusCode } = await downloadFileAndUnzip({
-		url: getWordPressVersionUrl(wordPressVersion),
+		url: getWordPressVersionUrl(
+			isDeveloperBuild ? 'nightly' : wordPressVersion
+		),
 		destinationFolder: tempFolder,
 		checkFinalPath: finalFolder,
 		itemName: `WordPress ${wordPressVersion}`,
--- a/packages/wp-now/src/tests/wp-now.spec.ts
+++ b/packages/wp-now/src/tests/wp-now.spec.ts
@@ -9,6 +9,7 @@ import {
 	isWpContentDirectory,
 	isWordPressDirectory,
 	isWordPressDevelopDirectory,
+	resolveWordPressVersion,
 } from '../wp-playground-wordpress';
 import {
 	downloadSqliteIntegrationPlugin,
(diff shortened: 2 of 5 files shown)


zaguiini:
cc @sejas

packages/wp-now/src/download.ts
```diff
@@ -37,15 +31,11 @@ function httpsGet(url: string, callback: Function) {
 	https.get(url, { agent }, callback);
 }
 
-function getWordPressVersionUrl(version = DEFAULT_WORDPRESS_VERSION) {
-	if (!isValidWordPressVersion(version)) {
-		throw new Error(
-			'Unrecognized WordPress version. Please use "latest", "trunk", "nightly", or numeric versions such as "6.2", "6.0.1", "6.2-beta1", or "6.2-RC1"'
-		);
-	}
```
sejas, line 45: I've found this check is already on:
https://github.com/zaguiini/playground-tools/blob/f8e1c1e15767ae0e391317afffbf61964e7fad7b/packages/wp-now/src/config.ts#L146-L150

sejas (commented):
@zaguiini , thank you for creating the PR. It looks good, working as expected.

I was thinking if it would be possible to remove `isDeveloperBuild` in favor of a more simple check, like directly using the provided alias.

Let me know if you want me to update the tests as well.

Thank you! 

I tested it and I confirm that after building the project `npx nx build wp-now`, I could run wp-now start with the following results:

`node dist/packages/wp-now/cli.js start`
`node dist/packages/wp-now/cli.js start --wp=latest`
It downloaded and used `~/.wp-now/wordpress-versions/6.5.5`

`node dist/packages/wp-now/cli.js start --wp=6.6-beta2`
It downloaded and used `~/.wp-now/wordpress-versions/6.6-beta2`

`node dist/packages/wp-now/cli.js start --wp=trunk`
`node dist/packages/wp-now/cli.js start --wp=nightly`
It downloaded and used `~/.wp-now/wordpress-versions/6.7-alpha-58716`
```
Starting the server......
directory: /Users/macbookpro/proyectos/a8c/playground-tools
mode: playground
php: 8.0
wp: 6.7-alpha-58716 (resolved from alias: trunk)
Downloading WordPress 6.7-alpha-58716...
Server running at http://localhost:8881
```






Passing an invented version returned a correct error:
```
node dist/packages/wp-now/cli.js start --wp=foo     

wp-now start

Start the server

Options:
      --version         Show version number                            [boolean]
      --path            Path to the PHP or WordPress project. Defaults to the cu
                        rrent working directory.                        [string]
      --php             PHP version to use.                             [string]
      --wp              WordPress version to use: e.g. '--wp=6.2'       [string]
      --port            Server port                                     [number]
      --blueprint       Path to a blueprint file to be executed         [string]
      --reset           Create a new project environment, destroying the old pro
                        ject environment.                              [boolean]
      --skip-browser    Do not launch the default browser
                                                      [boolean] [default: false]
      --inspect         Use Node debugging client.                      [number]
      --inspect-brk     Use Node debugging client. Break immediately on script e
                        xecution start.                                 [number]
      --trace-exit      Prints a stack trace whenever an environment is exited p
                        roactively, i.e. invoking process.exit().       [number]
      --trace-uncaught  Print stack traces for uncaught exceptions; usually, the
                         stack trace associated with the creation of an Error is
                         printed, whereas this makes Node.js also print the stac
                        k trace associated with throwing the value (which does n
                        ot need to be an Error instance).               [number]
      --trace-warnings  Print stack traces for process warnings (including depre
                        cations).                                       [number]
  -h, --help            Show help                                      [boolean]

Unrecognized WordPress version. Please use "latest" or numeric versions such as "6.2", "6.0.1", "6.2-beta1", or "6.2-RC1"
```

zaguiini:
>I was thinking if it would be possible to remove isDeveloperBuild in favor of a more simple check, like directly using the provided alias.

It's possible. I just wanted to stick as much as possible to the "a function should do one thing" principle. Also if we move the alias resolver within `downloadWordPress`, a bigger refactor will be required so that the "resolved from alias" output part shows before the version is downloaded.

But let me know if you have other ideas.

>Let me know if you want me to update the tests as well.

That'd be great 🙌

sejas:
I've fixed the tests here https://github.com/WordPress/playground-tools/pull/324/commits/e727b2ebed4c971e63b7638eea4f104e854e4d4f

@zaguiini, I've also realized that this PR was breaking the ability to use wp-now while offline. I used a file to cache the resolutions and use it in case the request to the server fails. Let me know what you think.

Test instructions:
- Run `npx nx build wp-now`
- Run `node dist/packages/wp-now/cli.js start`
- Observe a new file is created in the path `~/.wp-now/wp-version-offline-alias.json`
- The content of the file is similar to:
```
{
  "latest": "6.6.1"
}
```
- Kill wp-now by pressiong `ctrl + c`
- Disconnect your computer from internet
- Run again `node dist/packages/wp-now/cli.js start`
- Observe wp-now starts successfully 
- In the console appears a message: `Failed to fetch WordPress version online. Trying offline version.`

zaguiini:
Fantastic, @sejas!

All good to me. We're not handling the error when there is no downloaded corresponding WP version and it also fails to fetch from the internet:

image

But that's a small issue. Great job 🚀

sejas:
> All good to me. We're not handling the error when there is no downloaded corresponding WP version and it also fails to fetch from the internet:

Great suggestion. I've decided to terminate the process and displaying an error for that case: https://github.com/WordPress/playground-tools/pull/324/commits/aa9cde23414909c031f06bc47eef9ae40cdcc4e1

PR #326, opened 2024-07-11
Playground block: Add Open-in-New-Window feature

### LINKED ISSUE #279: Playground Block: Add a link to launch a full instance
Right now the block-viewing experience is limited to its iframe container. It'd be nice to have a user-facing link to launch a full window version.

brandonpayton:
## What?

Add ability to open a Playground block in a dedicated window.

### What it does

Opening in a new window reflects the current file content in the code editor as well as the activation status (i.e., if the iframe is already activated, it will be auto-activated when opening a new window).

### What it doesn't do

It doesn't relay WP data and state to the new window. Opening a new window starts with initial WP state.

Fixes #279

## Why?

Two reasons to add this are:
1. Users who rely on assistive technology have requested this because it is easier to interact with Playground as a dedicated, separate page.
2. Full-page Playground allows the user to work with both a larger code editor and a larger Playground iframe.

## How?

This PR:
- Adds support for a special query string to receive Playground block configuration and render it as a standalone page.
- Adds an "Open in New Window" link to the Playground block on the front end, when not already running in a full-page context.

## Testing Instructions

I've smoke tested with a variety of configuration, but I think the following are reasonable test instructions for review.

We need to test these configurations:
- Horizontal: Code editor side-by-side with preview - multiple files with error log
- Vertical: Code editor above preview - multiple files with error log
- No code editor

Test by creating a post containing one of each configuration and doing the following with each:
- Click the "Open in New Window" link and observe that the block contents are opened as a full-page in a new window.
- If there is a code editor, make some changes, open again in a new window, and observe that the code changes are reflected in the new full-page window.
- Activate the preview in the embedded block, click "Open in New Window", and observe that the new window auto-activates the preview.

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -135,7 +135,6 @@ export default function PlaygroundPreview({
 	const playgroundClientRef = useRef<PlaygroundClient | null>(null);
 	const fileMgrRef = useRef<FileManagerRef>(null);
 	const codeMirrorRef = useRef<any>(null);
-	const newWindowLinkRef = useRef<HTMLAnchorElement>(null);
 
 	/**
 	 * Prevent the CodeMirror keyboard shortcuts from leaking to the block editor.
@@ -311,12 +310,6 @@ export default function PlaygroundPreview({
 		return fullPageUrl.toString();
 	}
 
-	function freshenFullPageUrlOnClick() {
-		if (newWindowLinkRef.current) {
-			newWindowLinkRef.current.href = getFullPageUrl();
-		}
-	}
-
 	function getLandingPageUrl(postId: number = currentPostId) {
 		if (createNewPost && redirectToPost) {
 			if (redirectToPostType === 'front') {
@@ -414,16 +407,16 @@ export default function PlaygroundPreview({
 			>
 				<header className="demo-header">
 					{!inBlockEditor && !inFullPageView && (
-						<a
+						<Button
+							variant="link"
 							className="demo-header__full-page-link"
-							target="_blank"
-							ref={newWindowLinkRef}
-							href={getFullPageUrl()}
-							onClick={freshenFullPageUrlOnClick}
+							onClick={() => {
+								window.open(getFullPageUrl(), '_blank');
+							}}
 						>
 							Open in New Window
 							<span className="dashicons dashicons-external"></span>
-						</a>
+						</Button>
 					)}
 				</header>
 				<div className={mainContainerClass}>
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -301,11 +301,11 @@ export default function PlaygroundPreview({
 
 		const encodedFullPageAttributes = btoa(
 			JSON.stringify(base64EncodeBlockAttributes(fullPageAttributes))
-				);
+		);
 		fullPageUrl.searchParams.append(
 			'playground-attributes',
 			encodedFullPageAttributes
-				);
+		);
 
 		return fullPageUrl.toString();
 	}
@@ -392,8 +392,8 @@ export default function PlaygroundPreview({
 	const contentContainerClass = classnames(
 		'wordpress-playground-content-container',
 		{
-		'is-one-under-another': !codeEditorSideBySide,
-		'is-side-by-side': codeEditorSideBySide,
+			'is-one-under-another': !codeEditorSideBySide,
+			'is-side-by-side': codeEditorSideBySide,
 		}
 	);
 	const iframeCreationWarningForRunningCode = __(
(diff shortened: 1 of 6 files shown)


brandonpayton:
There are a couple of things left to do:
1. Fix the full-page URL to be simply based on site URL
2. Find a good place to put the Open in New Window link that is visible but out of the way. In addition, the link should be one of the first things encountered by users of assistive technologies so they don't have to navigate through the block before opening in a new window.

brandonpayton:
Here are some screenshots. Note that we are still looking for a good place for the "Open in New Window" link.

Front end:
Screenshot 2024-07-11 at 5 40 03 PM

Full page with side-by-side config:
Screenshot 2024-07-11 at 5 40 19 PM

Full page with top-bottom config:
Screenshot 2024-07-11 at 5 40 46 PM

Full page with no code editor:
Screenshot 2024-07-11 at 5 44 07 PM

brandonpayton:
IMO, it seems strange to launch full-page Playground with no code editor while still requiring activation. I wonder whether we should skip activation when launching as a full page. 🤔  I'll leave it as-is for now.

brandonpayton:
> 2. Find a good place to put the Open in New Window link that is visible but out of the way. In addition, the link should be one of the first things encountered by users of assistive technologies so they don't have to navigate through the block before opening in a new window.

I thought of putting a link as an overlay to the content but don't want it to possibly interfere with WordPress underneath. And we cannot make it part of the code editor because the code editor is not always present.

Currently, I am considering whether we can leave it as part of the section header but render it unobtrusively as part of the footer.

If you have another idea, please let me know.

brandonpayton:
As for general location in the layout, I think this may be a reasonable place for the link. If so, the styling certainly needs adjusted. The two items in the visual footer are not aligned with one another, and they don't look good together yet.

Screenshot 2024-07-11 at 6 02 50 PM

Perhaps we could shorten the visible text to "Full Page" with a helpful icon. The accessible text could remain more verbose.

brandonpayton:
Or perhaps the upper right is better if what is shown below can be made a bit lighter and less attention-grabbing:
Screenshot 2024-07-11 at 7 11 45 PM

brandonpayton:
The upper right location seems OK to me, but the text could be improved. @adamziel @bgrgicak what do you think?

https://github.com/user-attachments/assets/aee23be0-95f1-440c-8cc3-9f5cd46062e0

brandonpayton:
Next:
- Address PHP lint errors
- Write real testing instructions and make sure all options have been tested

Note: This PR is easier to review while ignoring whitespace:
https://github.com/WordPress/playground-tools/pull/326/files?diff=split&w=1

brandonpayton:
Fixed all lint errors and warnings except:
```
Processing form data without nonce verification.PHP_CodeSniffer(WordPress.Security.NonceVerification.Recommended)
```

🤔 I don't think we are worried about CSRF attacks here, but I'll sleep on it.

One thing I don't like is that we are rendering the block on the server side based on user input which seems like it could be risky in ways that are difficult to anticipate. Will think about this more as well.

brandonpayton:
> One thing I don't like is that we are rendering the block on the server side based on user input which seems like it could be risky in ways that are difficult to anticipate. Will think about this more as well.

Maybe we can:
- ignore the query params on the server-side
- render a basic Playground block that doesn't initialize until...
- deriving Playground block attributes from the query params on the client-side

brandonpayton:
I'm pivoting to the client-side route to avoid consuming query params on the server-side. This includes query params containing PHP code, and even though we only plan to relay that code back to the client-side, it is safest if the server does not touch it at all. In addition, relaying client-side to client-side will involve much less code and little formal query param interface.

adamziel:
> The upper right location seems OK to me, but the text could be improved. @adamziel @bgrgicak what do you think?

+1! I'd make it a WordPress `<Button>` for accessibility reasons, but I like that placement. What do you think @jarekmorawski?

adamziel:
> Maybe we can:
> 
> * ignore the query params on the server-side
> * render a basic Playground block that doesn't initialize until...
> * deriving Playground block attributes from the query params on the client-side

That sounds reasonable. Alternatively, the query params could be page ID and block ID and then the server would source data from an existing block rather than an rendering arbitrary block.

brandonpayton:
Thanks for the feedback!

I switched to only loading the full-page block attributes on the client-side, and it worked well immediately. It seems better to relay via query string rather than looking up by post and block ID because it is simpler and more flexible and doesn't require any DB lookups. In addition, it allows us to base the new full page view on a user's unpersisted changes to files within the block.

Looking at switching the open-new link to a button.

brandonpayton:
Thanks for catching that we should be using a button for accessibility purposes. That is now updated, and this is ready for review.

jarekmorawski:
Positioning the button above the preview frame will add an asymmetrical gap that may throw people off and make the block look unbalanced. I'd put it in the footer instead.

option1

Alternatively, we could connect it to the site title. Could something like this work?

option1-1

adamziel:
@brandonpayton I'm I'm not missing any accessibility-related nuance, the "Activate live preview" button doesn't seem to make much sense when opening the block in the new window.

I really like how it preserves the changes made in the code editor. :heart:

adamziel:
The button colors are slightly different when opening the block in the new window – how interesting! I don't see it as a blocker here, but it would be useful to track it in a separate issue.

packages/wordpress-playground-block/src/view.tsx
```diff
@@ -7,15 +7,40 @@ function renderPlaygroundPreview() {
...
+		urlParams.has('playground-full-page') &&
+		urlParams.has('playground-attributes') &&
+		playgroundDemo.length === 1
+	) {
+		const rootElement = playgroundDemo[0] as HTMLDivElement;
 		const root = createRoot(rootElement);
+		const encodedAttributes = urlParams.get(
+			'playground-attributes'
+		) as string;
+		const attributeJson = atob(encodedAttributes);
```
adamziel: Just to make sure – are there two layers of base64 encoding here? So the entire attributes string and each attribute value individually?

adamziel: `atob` and `btoa` won't correctly process UTF-8 characters, e.g.:

```js
> btoa('zażółć gęślą jaźń')
VM326:1 Uncaught DOMException: Failed to execute 'atob' on 'Window': The string to be decoded contains characters outside of the Latin1 range.
```

Instead, use `base64ToString` and `stringToBase64` exported from `base64.ts`:

```js
> stringToBase64( 'zażółć-gęślą-jaźń' );
'emHFvMOzxYLEhy1nxJnFm2zEhS1qYcW6xYQ='
```

adamziel: I went ahead and adjusted that in [4ee7a8c](https://github.com/WordPress/playground-tools/pull/326/commits/4ee7a8c953249d4373571f470f549341e471f28f)

adamziel (approved):
The code change is solid and works well so code-wise, I'm approving. About the design and the button placement, feel free to make the call based on what would be useful accessibility-wise @brandonpayton (since accessibility is the bulk of the motivation for this change).

packages/wordpress-playground-block/src/view.tsx
brandonpayton: > I went ahead and adjusted that in [4ee7a8c](https://github.com/WordPress/playground-tools/pull/326/commits/4ee7a8c953249d4373571f470f549341e471f28f)

Thanks for catching that and making the fix! ❤️

brandonpayton:
> @brandonpayton I'm I'm not missing any accessibility-related nuance, the "Activate live preview" button doesn't seem to make much sense when opening the block in the new window.

@alexstine When opening a Playground block in a dedicated window, is there any accessibility-related value to requiring the iframe to be activated in the full window before the preview iframe is loaded?

brandonpayton:
> The code change is solid and works well so code-wise, I'm approving. About the design and the button placement, feel free to make the call based on what would be useful accessibility-wise @brandonpayton (since accessibility is the bulk of the motivation for this change).

Great! I think we can probably place the open-in-new-window link in the header but style it to be rendered with the footer, using flexbox ordering and cross-axis properties. Will take a look.

brandonpayton:
> Great! I think we can probably place the open-in-new-window link in the header but style it to be rendered with the footer, using flexbox ordering and cross-axis properties. Will take a look.

It turned out to be doable using `flex-wrap`, `order` and `justify-content` properties, working with the main axis.

Something like this:
Screenshot 2024-07-15 at 7 06 47 PM

I don't love the spacing but think it might be OK for now. Will look at tweaking the styles a bit, but I think this is almost acceptable for the time being.

alexstine:
@brandonpayton 

> @alexstine When opening a Playground block in a dedicated window, is there any accessibility-related value to requiring the iframe to be activated in the full window before the preview iframe is loaded?

No.

Thanks.

brandonpayton:
This is close. I've eliminated the activation prompt when running as a full-page and have the open-in-new-window button placed first in the section header markup but laid out after the powered-by link in the footer:
Screenshot 2024-07-16 at 7 56 59 PM

The main thing left is to fix some incorrect positioning in the full-page vertical layout.

brandonpayton:
Fixed the remaining style issues.

Rebasing was super painful after the i18n PR touched so many lines, so I used `git rebase --ignore-whitespace`. This led to some whitespace-related lint violations, and I'm reviewing the rest of the changes for any other missing whitespace updates.

packages/wordpress-playground-block/src/components/playground-preview/index.tsx
```diff
@@ -634,11 +685,11 @@ export default function PlaygroundPreview({
 						}
 					</span>
 				</div>
-			</section>
-			<footer className="demo-footer">
+			</div>
+			<footer className="wordpress-playground-footer">
```
brandonpayton, line 689: Move the footer to be a child of the `<section>`. It was incorrectly outside of the section and needs to be part of the section for styling in this PR.

brandonpayton:
This is looking good. Since it is well past my EOD, let's wait to merge and tag a new release tomorrow.

PR #327, opened 2024-07-15
Playground block: Add i18n support

brandonpayton:
## What?

This PR adds i18n support to enable translation of Playground block strings.

## Why?

Before this PR, Playground block strings were not rendered using i18n facilities and were not translatable.

## How?

This PR uses `@wordpress/i18n`, adds a local i18n module with the correct text domain, and attempts to convert all fixed strings to using the i18n API.

## Testing Instructions

Smoke test. Use the block in the editor and on the front end. Play with different block settings toggles, and ensure the user visible text is intact.

@akirk Somehow it doesn't appear possible to select you as a reviewer when creating this PR. Would you be willing to review this as well? I am relatively new to i18n APIs and would love to learn from your wisdom here.

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -13,7 +13,12 @@ import {
 	phpVar,
 	// @ts-ignore
 } from 'https://playground.wordpress.net/client/index.js';
-import { useEffect, useRef, useState } from '@wordpress/element';
+import {
+	useEffect,
+	useRef,
+	useState,
+	createInterpolateElement,
+} from '@wordpress/element';
 import { Button, Spinner } from '@wordpress/components';
 import {
 	Icon,
@@ -34,6 +39,7 @@ import {
 	TranspilationFailure,
 	transpilePluginFiles,
 } from './transpile-plugin-files';
+import { __, _x, sprintf } from '../../i18n';
 
 export type PlaygroundDemoProps = Attributes & {
 	inBlockEditor: boolean;
@@ -87,7 +93,10 @@ export default function PlaygroundPreview({
 	logInUser,
 	createNewPost,
 	createNewPostType = 'post',
-	createNewPostTitle = 'New post',
+	createNewPostTitle = _x(
+		'New post',
+		'default title of new post created by blueprint'
+	),
 	createNewPostContent = '',
 	redirectToPost,
 	redirectToPostType = 'front',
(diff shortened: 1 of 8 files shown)


packages/wordpress-playground-block/src/components/playground-preview/index.tsx
```diff
@@ -390,8 +401,18 @@ export default function PlaygroundPreview({
 										}`}
 										aria-label={
 											isErrorLogFile(file)
-												? `Read-only file: ${file.name}`
-												: `File: ${file.name}`
+												? // translators: %s is a file name
+												  sprintf(
```
brandonpayton: I don't like how this text ended up being auto-formatted within the ternary expression. It might be better to assign to vars and just use them here for a more readable expression.

packages/wordpress-playground-block/src/components/playground-preview/index.tsx
```diff
@@ -594,11 +641,22 @@ export default function PlaygroundPreview({
...
 				>
-					<span className="demo-footer__powered">Powered by</span>
-					<Icon className="demo-footer__icon" icon={wordpress} />
-					<span className="demo-footer__link-text">
-						WordPress Playground
-					</span>
+					{createInterpolateElement(
+						// translators: powered-by label with embedded icon. please leave markup tags intact, including numbering.
+						__(
+							'<span1>Powered by</span1> <Icon /> <span2>WordPress Playground</span2>'
```
brandonpayton, line 647: The signature of `createInterpolateElement()` takes a map with keys for element names, so it is not possible to use regular `<span>` twice here and have them treated separately. But it turns out we can use unique non-existent tag names and achieve what we want.

It's not great because it might confuse translators who might "fix" our span "typos". To attempt to reduce the risk of this, I added a translators note that mentions the numbered markup tags.

packages/wordpress-playground-block/src/components/playground-preview/index.tsx
brandonpayton: Actually, these look fine in the editor but were just awkward during code review on GH. Will leave them as inline for now.

packages/wordpress-playground-block/src/i18n.ts
```diff
@@ -0,0 +1,9 @@
+import { createI18n, sprintf as coreSprintf } from '@wordpress/i18n';
+
+const i18n = createI18n(undefined, 'interactive-code-block');
```
bgrgicak, line 3: Nice, I didn't know that you could set the domain in one place instead of writing it on every line.

bgrgicak (approved):

packages/wordpress-playground-block/src/i18n.ts
brandonpayton: > Nice, I didn't know that you could set the domain in one place instead of writing it on every line.

I didn't at first either. This PR started with manually passing text domain everywhere (and discovering that I'd forgotten to add it in a bunch of places). This is much better. ☺️

Issue #316, opened 2024-06-19
Playground Block: A11y: Add text documentation for keyboard exit of CodeMirror


joedolson:
The `esc + tab` key combination for exiting the CodeMirror editor in the Playground block is supported, but not made evident to users. Having a visible text explanation would significantly improve the accessibility of the code view for keyboard users. 

Screen reader users have other ways of exiting the block, but keyboard users have limited options. 

Something to the effect of  "Use `esc` plus `tab` to tab forward out of the code editor." placed before entering the editor should be sufficient.

joedolson:
A slight clarification, since I realize that `esc + tab` technically means something different than the actual mechanism here. The process is to press `esc` and then press `tab` - pressing escape turns off the `tab` listener, so that the tab key will operate normally instead of being captured. It's a little bit different from the usual key combination pattern.

brandonpayton:
Thank you, @joedolson!

@jarekmorawski, we need to add some explanatory text so keyboard users know how to exit the code editor, and I do not currently have any great ideas about where to put it. Given the following UI screenshot, is there anything you might suggest?

![Screenshot 2024-07-23 at 12 09 11 PM](https://github.com/user-attachments/assets/977a40dc-d9a4-4e95-b4e3-29af52ffb73f)

jarekmorawski:
Could something like this work?

<img width="373" alt="image" src="https://github.com/user-attachments/assets/271ee843-7bf6-48a2-bf92-885531e64bb6">

joedolson:
It would be preferable if it was available to users before they entered the block. For low vision users, that text may be very difficult to find, since it may not be in their field of vision.  Additionally, if you've tabbed into the code editor while that region is outside of your viewport, it will be very difficult for you to scroll the viewport into a position where it's visible. E.g.

![image](https://github.com/user-attachments/assets/d2dc81ea-7410-41e9-9cd8-71cec95bf75d)

brandonpayton:
@jarekmorawski @joedolson what if there is a dismissible header between the file tabs and the code editor that explains how to exit the editor with the keyboard? If desired, we could use localStorage to remember it was dismissed so it isn't shown to the same user repeatedly (unless they are browsing in private mode where localStorage isn't persisted across sessions).

jarekmorawski:
Got it. Placing it outside the code editor feels counterintuitive because the block will display the whole Playground site and a button letting you close the code editor will be irrelevant when you're viewing the home page or managing plugins.

[...]

PR #340, opened 2024-07-29
Playground block: Add dismissible tip for exiting the editor with the keyboard

brandonpayton:
## What?

This PR adds a dismissible tip at the top of the editor to help keyboard-centric users know how they can leave the editor once it is focused.

Related to #316 

## Why?

As @joedolson [said](https://github.com/WordPress/playground-tools/issues/316):
"Having a visible text explanation would significantly improve the accessibility of the code view for keyboard users."

## Testing Instructions

- Run `npx nx dev wordpress-playground-block`
- View a post that has the Playground block
- Note the notice at the top of the editor
- Click "Dismiss" to dismiss the notice
- Confirm the notice is hidden
- Refresh the page and confirm the notice is no longer shown

To repeat this, open the dev tools and delete the key "playground-block-dismiss-exit-editor-tip".

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -164,6 +164,26 @@ export default function PlaygroundPreview({
 	);
 	const [currentPostId, setCurrentPostId] = useState(0);
 
+	const dismissedExitWithKeyboardTipKey =
+		'playground-block-dismiss-exit-editor-tip';
+	const [dismissedExitWithKeyboardTip, setDismissedExitWithKeyboardTip] =
+		useState(localStorage[dismissedExitWithKeyboardTipKey] === 'true');
+	function dismissExitWithKeyboardTip() {
+		localStorage[dismissedExitWithKeyboardTipKey] = 'true';
+		setDismissedExitWithKeyboardTip(true);
+
+		// Shift focus to editor so focus is not lost as the tip disappears
+		if (codeMirrorRef?.current?.view?.dom) {
+			const contentEditableElement: HTMLElement =
+				codeMirrorRef.current.view.dom.querySelector(
+					'[contenteditable=true]'
+				);
+			if (contentEditableElement) {
+				contentEditableElement.focus();
+			}
+		}
+	}
+
 	/**
 	 * Let the parent component know when the state changes.
 	 */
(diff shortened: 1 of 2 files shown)


brandonpayton:
@joedolson, what do you think of this? The notice is focusable and dismissible with Enter key or a click.

Screenshot 2024-07-29 at 7 11 13 PM

brandonpayton:
I think this notice probably needs an ARIA role since it is just a div.

bgrgicak (approved):

joedolson:
@brandonpayton I don't think there's any particularly need for an ARIA role on this; it's just page content. There should be a spoken announcement to confirm that the notice is dismissed, however.

joedolson:
I see; the dismiss 'button' is actually just a div. Is there a reason you're not actually using a button?

brandonpayton:
@joedolson You were right. I tried to simplify this and just used a button for the entire notice.

> There should be a spoken announcement to confirm that the notice is dismissed, however.

I started looking at this but could use some advice because nothing I've tried so far is leading to right result. (I am testing with macOS VoiceOver by default but plan to check in NVDA as well).

What would be the cleanest way to announce the notice is dismissed? Should this be doable with using `aria-live` and related attributes?

joedolson:
I think that having the entire panel be a button has some benefits, in that it guarantees that a user won't miss the instructions. 

If the 'Dismiss' text looks like it's clickable, however, it will be disconcerting to keyboard users who will not be able to focus it - and there shouldn't be focusable controls nested inside another control element. So perhaps that should have a design that makes it clear that it's the entire component that's reactive.

Here's some docs on the `withSpokenMessages` component: https://developer.wordpress.org/block-editor/reference-guides/components/with-spoken-messages/

brandonpayton:
Thank you for the link to "withSpokenMessages"! I didn't know about that.

I wonder if it would be better to just be direct with an additional "(Click to dismiss this notice.)" in the banner like:
Screenshot 2024-07-30 at 3 49 13 PM

brandonpayton:
I'm currently struggling to get the notice to be spoken by VoiceOver. Sometimes it has worked after VoiceOver finishes reading the entire contents of the code editor, which is focused as the keyboard-related is being dismissed. I tried using `speak(message, "assertive")`, but so far, that doesn't seem to result in anything being spoken.

brandonpayton:
@joedolson when the notice is dismissed, I am moving the focus to the code editor so the focus is not lost. Is it right to say we need this kind of focus management when removing an element that was previously focused? Changing focus to the editor seemed like a meaningful focus transition since the user is acknowledging how to exit the editor.

If focusing the editor on dismissal is the right course, what I am finding is that the focus change interferes with the spoken announcement. In my testing, if the focus change is made in response to dismissing the exit-editor notice, VoiceOver does not speak an assertive message. It will speak a polite message but only after reading what appears to be the entire content of the editor, making the message untimely.

Do you have any suggestions about how we could address this?

brandonpayton:
In order to keep things moving, I am going to merge this PR as is and create a follow up issue to address the spoken dismissal notification.

brandonpayton:
The follow-up issue for the spoken announcement is https://github.com/WordPress/playground-tools/issues/343.

PR #345, opened 2024-08-01
WIP: Update playground dependencies to v0.9.X

sejas:
## What?

Let's update Playground dependencies to the latest version.

We may need to update other projects like Playground Plugin.

## Why?

It will fix WBEP images, support curl and other fixes features introduced in these versions.

## How?

- We need to update the dependencies and adapt the breaking changes to use the new functions.

## Testing Instructions

- Run `npx nx build wp-now && node dist/packages/wp-now/cli.js start`
- Observe Playground opens correctly in the browser.


swissspidy:
Would love to be able to use the new version via wp-now 😊

eliot-akira:
I wanted to contribute to this pull request, and have started looking at what's needed to update Playground dependencies from `0.6.16` (current) to `0.9.31` (newest). I can try working on this if you could advise/review? @sejas @adamziel 

---

After installing the new versions, there are 29 failing tests. From the error messages and studying the code/tests of PHP-WASM modules, I learned a few things.

- Use `PHP` class from `@php-wasm/universal` instead of `NodePHP` from `@php-wasm/node`, which no longer exists. WordPress/wordpress-playground#1457
- Use `PHPRequestHandler` and let it manage PHP instances
- Use `createNodeFsMountHandler` to mount the file system with Node-specific adaptor

My fork of this PR branch shows the progress so far.

https://github.com/WordPress/playground-tools/compare/update/playground-dependencies...eliot-akira:playground-tools:update/playground-dependencies

(I wasn't sure if I should create a new pull request, or if this current PR can have multiple authors. I'm guessing the latter is not possible since I don't have write access to the PR branch, [update/playground-dependencies](https://github.com/WordPress/playground-tools/tree/update/playground-dependencies).)

---

The tests run a little further now (25 fails) but `php.mount()` is not working yet. It throws:

```
ErrnoError {
  name: 'ErrnoError',
  errno: 44,
  stack: '<generic error, no stack>'
}
```

One thing I noticed is, the `php.mount()` method is async ([defined here](https://github.com/WordPress/wordpress-playground/blob/0fd051ac5bd566ada407f54c33c638babffca61d/packages/php-wasm/universal/src/lib/php.ts#L998-L1001) in `@php-wasm/universal`) and I'm awaiting it - but elsewhere it's being used, like [in this test](https://github.com/WordPress/wordpress-playground/blob/0fd051ac5bd566ada407f54c33c638babffca61d/packages/php-wasm/node/src/test/php.spec.ts#L892) for `@php-wasm/node`, it's always called without `await`.

---

There are a couple places where the code was previously applying a callback to all PHP instances, but the new `PHPRequestHandler.processManager` doesn't provide access to them. [`PHPProcessManager.allInstances`](https://github.com/WordPress/wordpress-playground/blob/0fd051ac5bd566ada407f54c33c638babffca61d/packages/php-wasm/universal/src/lib/php-process-manager.ts#L78-L82) is a private property, probably for a good reason - maybe it's no longer necessary (or possible) to manage the instances manually.

---

During the process I learned about Playground CLI (WordPress/wordpress-playground#1289) and the new Boot Protocol (WordPress/wordpress-playground#1390). I can see these are generalizing and modularizing some features of `wp-now`, as well as parts of the build process for the Playground itself.

The [`@wp-playground/cli`](https://github.com/WordPress/wordpress-playground/tree/trunk/packages/playground/cli) module and [`bootWordPress()`](https://github.com/WordPress/wordpress-playground/blob/trunk/packages/playground/wordpress/src/boot.ts#L112) are good up-to-date references of how to use the `PHP` class properly. I'm comparing it with `wp-now` to understand what needs to be updated.

adamziel:
@eliot-akira AFAIR you got all the major changes right. Here's a major non-breaking change that could help with managing workers:

https://github.com/WordPress/wordpress-playground/pull/1669

eliot-akira:
Thank you, I'll keep chipping away at this. I've continued studying the source of `@php-wasm/node`, `universal`, and `@wp-playground/wordpress`. I think I can eventually figure out how to update `wp-now` and pass all existing tests.

eliot-akira:
OK, I got it down to 3 failing tests.

I created a new pull request from my forked branch, since the scope is narrower than this one (update `wp-now` only instead of all packages in `playground-tools`) and I wanted to show the test error messages.

- #350

sejas:
Great work @eliot-akira ! 🙌 , thanks for helping on this. I'm back from holidays and catching up. I'll review the [PR](https://github.com/WordPress/playground-tools/pull/350) in the next days.
The team has also updated the dependencies for Studio: https://github.com/Automattic/studio/pull/491

PR #346, opened 2024-08-09
Playground block: Fix visual tab order of open-in-new-tab button

### LINKED ISSUE #344: Playground block: open-in-new-tab focus order is visually out-of-sequence
From https://github.com/WordPress/Learn/issues/2783

I thought it would be helpful for the open-in-new-tab link to come first in the focus order for those using assistive technologies, even though we wanted to display the link in the block footer. But it naturally turns out to make visual focus order confusing for keyboard-centric users.

brandonpayton:
## What, why, and how?

The open-in-new-tab button is visually positioned in the block footer, but in the markup, it was originally placed in the section header so the link would be the first thing focused by those using assistive technologies. Unfortunately, that leads to a confusing visual tab order because a link in the block footer is focused first, followed by everything else.

This PR places the open-in-new-tab button in the footer, so the tab order matches the intuitive visual order.

Fixes #344

## Testing Instructions

- Run `npx nx dev wordpress-playground-block`
- View a post with a Playground block
- Confirm that the button remains in the footer with a reasonable location and spacing
- Click on the button to open the block in a full window
- Confirm that the open-in-new-tab button is hidden in the full window view

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -433,26 +433,6 @@ export default function PlaygroundPreview({
 			aria-label={__('WordPress Playground')}
 			className={mainContainerClass}
 		>
-			<header className="wordpress-playground-header">
-				{!inBlockEditor && !inFullPageView && (
-					<Button
-						variant="link"
-						className="wordpress-playground-header__full-page-link"
-						onClick={() => {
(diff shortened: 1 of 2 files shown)


brandonpayton:
After the update:
Screenshot 2024-08-09 at 3 13 13 PM

brandonpayton:
There was some lost vertical spacing between the footer and the block. That is fixed now.
Screenshot 2024-08-09 at 3 31 18 PM

Issue #343, opened 2024-07-30
Playground block: Dismissing exit-editor-with-keyboard tip should trigger a spoken announcement


brandonpayton:
When the user dismisses the exit-editor-with-keyboard tip, there should be a spoken announcement.

This is a follow up for PR #340 where there were difficulties getting the announcement to be spoken in a timely manner. 
See https://github.com/WordPress/playground-tools/pull/340#issuecomment-2259309390

brandonpayton:
Quoting from [this comment](https://github.com/WordPress/playground-tools/pull/340#issuecomment-2259309390):
> @joedolson when the notice is dismissed, I am moving the focus to the code editor so the focus is not lost. Is it right to say we need this kind of focus management when removing an element that was previously focused? Changing focus to the editor seemed like a meaningful focus transition since the user is acknowledging how to exit the editor.
> 
> If focusing the editor on dismissal is the right course, what I am finding is that the focus change interferes with the spoken announcement. In my testing, if the focus change is made in response to dismissing the exit-editor notice, VoiceOver does not speak an assertive message. It will speak a polite message but only after reading what appears to be the entire content of the editor, making the message untimely.
> 
> Do you have any suggestions about how we could address this?

The fact that the entire editor text seems to be read on focus is one problem. The focus change to the editor seems to interrupt an assertive message, and if we use a polite message, we may wait a very long time for the editor contents to be read before a spoken notice of banner dismissal.

As a compromise, I wonder if we can set the focus to the active file tab instead. Then, even if we have to wait for it to be read, the spoken message will at least be timelier than a notice read after all editor contents.

brandonpayton:
@joedolson, when the notice is dismissed and removed from the DOM, we move the focus to the code editor so focus is not lost but rather moved to a relevant place following the dismissal. Unfortunately, the fact that we move focus to the editor causes VoiceOver to read the contents of the editor (which can be substantial) before reading a polite notice that the notice has been dismissed. If we make the spoken notice assertive, the change in focus seems to completely skip the reading of the assertive notice.

Perhaps other screen readers perform better in this case.

Regardless, I've tried various approaches that would change the order and timing of the the focus and the speaking but  haven't been able to figure out a good way to get have dismissal spoken by VoiceOver in a timely matter.

Do you have any suggestions? It seems like the need to speak and the need to transfer focus are at odds with one another.

The notice itself says "Click to dismiss". Does that make this circumstance any more tolerable if we cannot effectively speak the dismissal?

[...]

PR #348, opened 2024-08-19
Playground block: a11y: Speak notice dismissal

brandonpayton:
## What and why?

This PR adds a spoken notice for those using assistive technologies when the exit-the-editor-with-keyboard notice is dismissed.

Related to #343

## How?

The Playground block is wrapped using `withSpokenMessages()` from `@wordpress/components`. The spoken message is "polite" and does not interrupt the screen reader.

Prior to this PR, the editor itself was focused when the exit-with-keyboard notice was dismissed, but this was not good for two reasons:
1. It triggers reading editor content which is potentially very long.
2. When the editor text is being read, the user will have to wait a long time for the "Dismissed notice" message to be spoken.

This PR now focuses the previous item in the tab order (the download button) when the exit-with-keyboard notice is dismissed.

## Testing Instructions

- Enable a screen reader
- Run npx nx dev wordpress-playground-block
- View a post that has the Playground block
- Move focus to the exit-with-keyboard notice at the top of the editor
- Click the notice to dismiss it
- Confirm that "Notice dismissed" is spoken after the notice is dismissed. This may happen after the download button text is read.

To repeat this, open the dev tools and delete the local storage key "playground-block-dismiss-exit-editor-tip" and refresh the page.

### DIFF
--- a/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
+++ b/packages/wordpress-playground-block/src/components/playground-preview/index.tsx
@@ -19,7 +19,7 @@ import {
 	useState,
 	createInterpolateElement,
 } from '@wordpress/element';
-import { Button, Spinner } from '@wordpress/components';
+import { Button, Spinner, withSpokenMessages } from '@wordpress/components';
 import {
 	Icon,
 	plus,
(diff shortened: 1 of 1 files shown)


brandonpayton:
This has been tested with macOS VoiceOver. Due to trouble with my Windows VM this morning, this was not tested with NVDA.

PR #354, opened 2024-09-06
Update contributing instructions to include wp-now installation

jonathanbossenger:
## What?

Adds a step to install wp-now globally, which powers the local development environment for contributing to the playground block.

## Why?

In setting up the local development environment for the playground block, I discovered I needed wp-now installed.


adamziel:
Thank you @jonathanbossenger!

Issue #290, opened 2024-06-04
Playground block: A11Y: Difference between "Run" and "Activate Live Preview"?


alexstine:
Test page: https://learn.wordpress.org/test/wordpress-playground-block-plugin-test-page/

From the button labels alone, I cannot tell the difference between what these buttons will do. Let's start with answering that question and then I can guide on the proper solution. This is in the preview iFrame.

Run sounds like it would load the iFrame from the provided code but I have no idea what Activate Live Preview would mean. From my perspective, they appear to do the same thing.

brandonpayton:
Hi @alexstine, thanks for bringing this up.

> Run sounds like it would load the iFrame from the provided code but I have no idea what Activate Live Preview would mean. From my perspective, they appear to do the same thing.

Before live preview has been activated, both the "Run" and "Activate Live Preview" buttons have different purposes but effectively do the same. The "Run" button serves to run the code editor's content in Playground, but when the live preview is not yet active, "Run" has to activate the preview in order to run the code. The "Activate Live Preview" button serves to communicate that user action is required to load the live preview, and when activating the preview, it run the code from the editor as a side-effect.

After live preview is activated, the "Run" button remains and serves to run the latest contents of the code editor in Playground.

Does that help?

alexstine:
@brandonpayton Any chance you can remove the Activate live preview button? Run will ultimately do the same thing and it will be far less confusing.

brandonpayton:
@alexstine I feel like the "Activate live preview" button may have value from a visual UX perspective because it is centered on the live preview area while the "Run" button is located separately with the code editor. When using a screen reader, encountering a "Run and Activate" button followed by a separate "Activate and Run" button sounds confusing and like it would cause a person to pause and wonder "What's the point? Am I missing something?"

Is there a good way for the "Activate live preview" button to be excluded from screen reader consideration to avoid this?

alexstine:
@brandonpayton Nope, that's not possible. After selecting activate live preview, it also gets removed from the DOM which is also not a great experience.

brandonpayton:
> @brandonpayton Nope, that's not possible. After selecting activate live preview, it also gets removed from the DOM which is also not a great experience.

@alexstine I do not believe we should remove the "Activate live preview" button because there is also a Playground block mode where there is no code editor and it is the only button that can activate the preview. And it seems good to keep the Preview UI consistent between the code-editor and no-code-editor modes.

[...]

PR #355, opened 2024-09-07
Playground block: Reflect preview loading status for screen readers

brandonpayton:
## What and why?

This PR adds a couple of measures to communicate Playground preview loading status:
- Updating the before-preview marker to say one of the following depending on state:
  - "Beginning of Playground Preview - Not activated"
  - "Beginning of Playground Preview - Loading"
  - "Beginning of Playground Preview - Loaded"
- Speaking "WordPress Playground loaded" after Playground completes loading

Related to #290 

## Why?

We received a11y feedback that the preview loading status is not clearly communicated.

## Testing Instructions

- Start a screen reader
- View a post containing a Playground block that requires activation
- Click the Activate button
- Observe that the before-preview marker is focused and that its text updates in response to preview loading status
- Observe that "WordPress Playground loaded" is spoken once the preview completes loading

### DIFF
--- a/packages/wordpress-playground-block/src/style.scss
+++ b/packages/wordpress-playground-block/src/style.scss
@@ -121,7 +121,7 @@
 			}
 			.playground-container {
 				width: 100%;
-				height: 400px;
+				height: 500px;
 				flex-basis: 500px;
 			}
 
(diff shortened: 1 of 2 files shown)


packages/wordpress-playground-block/src/style.scss
```diff
@@ -121,7 +121,7 @@
 			}
 			.playground-container {
 				width: 100%;
-				height: 400px;
+				height: 500px;
```
brandonpayton, line 124: This is an unrelated style fix.

PR #350, opened 2024-08-27
wp-now: Update Playground and PHP WASM dependencies

eliot-akira:
## What?

This pull request is a continuation of #345 to update Playground dependencies from `0.6.16` to `1.0.2` and adapt existing code to work with breaking changes.

Resolves #319.

## Why?

To keep up with upstream changes and benefit from new features and bug fixes.

## How?

Update to new interface for PHP runtime instances, requests, and file system mount.

Based on similar logic in `@wp-playground/wordpress` ([`bootWordpress`](https://github.com/WordPress/wordpress-playground/blob/56502a4027c3585d3c13c95f21300fd3ddd3589a/packages/playground/wordpress/src/boot.ts#L96-L112)) and `cli`.

- Use `PHP` class from `@php-wasm/universal` instead of `NodePHP` from `@php-wasm/node`, which no longer exists.
  - WordPress/wordpress-playground#1457
- Use `PHPRequestHandler` instead of `php.request()`
- Use `createNodeFsMountHandler` to mount the file system with Node-specific adaptor for `php.mount()`
- Remove logic that handled PHP instances manually

Further changes were necessary to pass all existing tests.

- Update `executeWpCli` based on [fork of `wp-now` in Automattic/studio](https://github.com/Automattic/studio/blob/trunk/vendor/wp-now/src/execute-wp-cli.ts)
  - Use TLS certificates
  - Workaround for stdin/out/err
  - Not implemented yet: SQLite command
- ~Update `@php-wasm/compile` to add any missing C functions to the list in `ASYNCIFY_IMPORTS` in the Dockerfile [here](https://github.com/WordPress/wordpress-playground/blob/ca0991a2d31fbf7faa82c3f3b7acac74866a8e20/packages/php-wasm/compile/php/Dockerfile#L439).~
  - Documentation about WASM errors related to Asyncify: https://wordpress.github.io/wordpress-playground/developers/architecture/wasm-asyncify/#asyncify-crashes

## Testing Instructions

- `npx nx test wp-now`

PR #359, opened 2024-10-17
fix vscode-extension bundle step adding correct jspi subfolder

sejas:
## What?

It updates the correct path of `jspi/PHP_*` to be included in the VS Code extension after upgrading the Playground dependencies on https://github.com/WordPress/playground-tools/pull/350

## Why?

The paths to wasm files have changed in the new version.

## How?

Updating the project bundle step.

## Testing Instructions

* Run `npx nx run vscode-extension:build:bundle`
* Observe the command finishes successfully : ` >  NX   Successfully ran target build:bundle for project vscode-extension (501ms)`

### DIFF
--- a/packages/vscode-extension/project.json
+++ b/packages/vscode-extension/project.json
@@ -46,8 +46,8 @@
 					"cp ./LICENSE dist/packages/vscode-extension",
 					"cp packages/vscode-extension/package.json dist/packages/vscode-extension",
 					"cp packages/vscode-extension/README.md dist/packages/vscode-extension",
-					"cp -r node_modules/@php-wasm/node/7_* dist/packages/vscode-extension",
-					"cp -r node_modules/@php-wasm/node/8_* dist/packages/vscode-extension"
+					"cp -r node_modules/@php-wasm/node/jspi/7_* dist/packages/vscode-extension",
+					"cp -r node_modules/@php-wasm/node/jspi/8_* dist/packages/vscode-extension"
 				],
 				"parallel": false
 			}


sejas:
After upgrading the playground dependencies, the bundle steps for `vscode-extension` and `interactive-code-block` started failing.

In this PR I'm fixing the `vscode-extension` bundle step, but to fix `interactive-code-block` it seems we need to make the change in the wordpress-palyground repository. Here is the error I'm seeing:

web-php-jspi-wasm

It seems it was modified in this commit:
https://github.com/WordPress/wordpress-playground/commit/3bebb247c75b4938fb6addc36abd1f19fe54cf1f

@adamziel , do you know if it's a fail in the import or in the "build" process?
I'm seeing the import is correct on 
https://github.com/WordPress/wordpress-playground/blob/2e48fa72a18ea222d697ff0602d3694f74b9162d/packages/php-wasm/web/src/lib/get-php-loader-module.ts#L22

So I'm wondering if we need to change the build process on the Playground repo to avoid having that php folder there.

adamziel:
Good catch @sejas, `@php-wasm/web` needs a fix to produce the correct path (or flatten the php folder in the built package). CC @psrpinto – would you like to dive in and get some more exposure to Playground's build tooling?

psrpinto:
@adamziel Absolutely, I'd be happy to look into it. On it.

psrpinto:
I created a fix for the issue at https://github.com/WordPress/wordpress-playground/pull/1958

PR #360, opened 2024-10-24
Update README.md to correct typo in extension path

jasonbahl:
## What?

Updating path in readme to match actual path in repo

## Why?

Typo in readme that didn't match actual path to code

## Testing Instructions

1. Set up the extension on chrome browser following the corrected instructions
2. verify that the UI in Github is now visible

CleanShot 2024-10-24 at 10 15 50


bgrgicak (approved):
Good catch @jasonbahl, thank you for contributing!

